Issue #31, opened 2017-11-28
WARN civic@ No repository field, No license field during npm install


MikeTheCanuck:
At the end of the `npm install` sequence, the following two warnings are displayed:

```
npm WARN civic@ No repository field.
npm WARN civic@ No license field.
```

DavideDaniel:
Not all Npm warnings are actual bugs and will be largely unavoidable with secondary dependencies.

jaronheard:
#42 addressed

PR #40, opened 2017-12-28
Add project metadata

MikeTheCanuck:
Added project metadata that is common to all child projects, and a description captured from civicpdx.org.

Inserted "@hackoregon" prefix in the `name` field based on how naming was done in the child projects.

This will explicitly address #31


MikeTheCanuck:
With no comments or criticism in over two months, and being a metadata-only PR, I've decided to merge this.

PR #42, opened 2018-04-14
Civic V2

DingoEatingFuzz:
To see it in action:

```sh
$ yarn bootstrap
$ yarn build
$ cd packages/2017
$ yarn run start:prod
```

### The overarching goal of these changes is to follow through with the plan of last year. In short, that means:

1. **DRYing out the webpack/babel/node server/scripts in each project**
  Last year each project had identical copies of express servers, webpack configs, babel presets, and other tools that slowly began to drift as the season progressed. In addition to the natural drift, when we made platform-wide changes to this tooling, it required 5+ pull requests. By extracting as much of it out into common dependencies, now the dependency gets updated, all dependent packages get tested, and a small changeset gets merged.

2. **Assembling all projects in a single React app that can be deployed as one unit**
  The story at the beginning of the 2017 project cycle was we were all working on individual packages that would eventually plug into a larger package that would actually get deployed. Even though we never quite got there last year, the idea was still very good. Now, as the commands above allude to, there is a `2017` package that does exactly this. It is one express server that serves one React code base that includes all routes, reducers, and components from each 2017 project and the platform package that had the home page and about page.

### I have some goals of my own for this year that I also kept in mind while making changes:

1. **No one should have to think about tooling in order to make story cards**
  Last year we lost a lot of time and energy to what I call meta-work: work that you do so you can do work. This meant communicating webpack changes, figuring out how to get Leaflet's CSS loaded, attempting to help other teams but having to first get familiar with their slightly different configurations. 
  Ultimately these are all important questions to answer and solve for, but they can kill productivity. By DRYing out the tooling, a lot of this headache will go away, but I made a couple decisions that go beyond this and further improve developer experience. They're outlined later.

2. **The project should already be a deployable website before any frontend devs for the 2018 project cycle starting making changes.**
  This PR doesn't quite get us to a deployable state, but it's almost there. A Travis build matrix is in place, and the 2017 package has a working production mode. All that's left is a containerization and deploy-to-ecs story. A 2018 package will be quick to follow this. The package will be more or less empty, since the projects have no story cards, but it will mean end-to-end development and deployment will be in place.
3. **No code or packages should go unused or unexplained.**
  There was a lot of code that never had documentation; even simple READMEs. There was also a lot of code that was committed to master in a WIP state. Meaning no explanation, not used by other packages, and plenty of commented out code making even the ambitious code spelunker left wondering what is going on. To this end, a lot of packages were completely removed and all new packages have READMEs.

### New Packages

  1. dev-server
  2. mock-wrapper
  3. webpack-common
  4. 2017

In summary, 2017 is the new `platform` package, and the other three are utility packages created to DRY out common tooling.

### Removed Packages

  1. civic-logger (unused)
  2. civic-scripts (unused)
  3. platform (replaced by 2017)
  4. webpacker (replaced by webpack-common)

The spirit of platform and webpacker are carried on by other packages. Civic Logger never got used, but the dev-server package does have some logging. As for Civic Scripts, it was never used, but it was _a lot_ of code. It looked like the goal was to eventually have a `civic` cli for easily working with projects. This is a novel idea, but I don't see anyone building it, and I don't like having unused code sitting around. What is currently the HEAD of the master branch has been tagged as `2017`, so if we want to dig this back out at a later date, it will be easy to extract.

### Developer Experience

To make development smoother and less surprising, I added some tools and rewrote some existing code.

The first big change is getting rid of CSS Modules in favor of [Emotion](https://emotion.sh/). CSS Modules is pretty much all or nothing, and this caused headaches and confusion last year. The workaround ended up putting loader directives in our imports, which was just hacky. For awhile I played with using regular expressions like webpacker had, but this felt brittle. Instead, the goal of CSS Modules (CSS encapsulation and namespacing) is still met with Emotion, a CSS-in-JS library that is authored a lot like Sass.

Last year, most developers wanted to write plain CSS. This works as expected now, but it still needs to be exercised with caution, since each project will be coexisting with four other projects.

Another DX tool added is one that some projects added on their own last year: a redux logger. This logs all actions and payloads to the console as they are emitted, which greatly eases in the process of creating reducers, selectors, and components.

### Going Forward

There is a lot more that can be done, but at this point I'm happy calling it a V2 and inviting more people to work on it. It is now in a state where I believe a developer experienced with React and Redux could self-orient and make meaningful contributions.

Some immediate next steps are

  1. Create a 2018 package
  2. Create packages for each 2018 project
  3. Dockerize the 2017 package
  4. Deploy the 2017 container, the 2018 container, and storybook as part of Travis
  5. Give Component Library some attention and embellishment 

If you made it this far, thanks for reading 🎉


jaronheard:
Ahhhhhhhhhh so exciting! Yes!

jaronheard:
Who gets the pleasure of reviewing and merging this? Technically, I could do it. But providing a meaningful review is probably beyond me.

MikeTheCanuck:
One question though, for my benefit surely but also for those folks who might work with this repo casually like me:

Why is it that - when I've gone through the full specification of commands (use/bootstrap/build/start), then `git pull` changes for both `master` and the current branch - I sometimes find myself experiencing errors that go away if I `rm -rf` my copy of the repo, clone it again, and run through the same commands?

Is there some command I could use before running through bootstrap/build/start that would clean up whatever mess causes these collisions?

And if this is not unheard of for others, can we find some way to integrate that command/troubleshooting step into the README?

DingoEatingFuzz:
Thanks for checking out all the packages, @MikeTheCanuck. The `createLogger` error I should be able to fix. It must have been introduced late when I started adding dev tools to the utility packages.

The `cannot read of Symbol of undefined` error is one that I can't seem to shake, but it's both intermittent (shows up ~85% of the time) and goes away with incremental builds, (i.e., if you save some component file, a rebuild is triggered, and the page loads as expected).

> Why is it that - when I've gone through the full specification of commands (use/bootstrap/build/start), then git pull changes for both master and the current branch - I sometimes find myself experiencing errors that go away if I rm -rf my copy of the repo, clone it again, and run through the same commands?

Some of the issues aren't consistent, so it has nothing to do with re-cloning. It's just a race condition somewhere.

DingoEatingFuzz:
Okay. @MikeTheCanuck, I addressed both the `createLogger` issue and the `Symbol of undefined` issue.

meganmckissack:
Amazing!  I ran everything via the instructions and yarn builds were successful! The only issues I ran across:

1. The above mentioned 'symbol of undefined' error showed up in the console while running the dev version of the budget package, causing nothing to show up on the page.  Running the prod server on the budget package worked perfectly.

2.  Storybook ran perfectly but looks like it's missing some elements (ie recharts graphs and some maps), these missing components did show up during the build logging.  Could be an npm versioning issue?

Running the production server on the 2017 package had everything showing up great! With headers and links even! So exciting.

Awesome, work Michael!  I'm in total agreement with the way you've structured the packages so that we're not dealing with so many unused dependencies and updates, there's lots to improve on the old packages but it so much easier to do that now.

I am so down to approve this PR and update package readmes later.  
:squirrel:

DingoEatingFuzz:
I spent an embarrassing number of hours today digging into the `Symbol of undefined` error and came up short.

What I learned was that it only happens in dev builds and it only happens when webpack decided to split code into multiple bundles. On my machine, webpack doesn't have deterministic behavior and will sometimes elect not to code split, which made this even harder to debug. When the error does arise, it's because webpack has determined that the module `lodash-es/_root` is not used by any code and can be marked as such. However, `lodash-es/Symbol` uses it and attempts to read a property off the undefined dependency.

The reason we were only seeing this error in certain packages despite all packages having identical webpack configs and dev servers is not all packages had multiple pages. Packages that don't have multiple packages never code split and never get into this situation.

Despite all this learning, I wasn't able to identify and address the root cause. I was however able to come up with a workaround.

Now, after successfully compiling the webpack config, the dev server will force an immediate rebuild. This resolves the broken module and any site loads as expected.

If someone else can confirm that all packages start, I'm comfortable merging this.

meganmckissack:
One question before merging, now that you've changed webpack to build production files to a dist folder instead of a build folder, will this trigger any errors or overwrites on the previous year's packages when we merge this pr?  Each package has a travis.yml file to trigger the npm run build script and deploy whatever's in the build folder to an s3 bucket after we merge to the master branch, which is how the current civic site is hosted.  Seems like nothing should happen after your changes but I wanted to check in and make sure.

DingoEatingFuzz:
Not only will it not error or overwrite previous S3 contents, those per-package travis files do nothing since they aren't at the repo root.

In the event that we need to change those files, we would have to make changes to the legacy individual repos.

I'll quickly delete those travis files, then merge this.

Thank you @meganmckissack and @MikeTheCanuck for testing out the changes!

Issue #48, opened 2018-04-23
Convert line chart to VictoryCharts


jaronheard:
We are moving from Recharts to VictoryCharts. See also issue #44.

The existing LineChart component needs to be converted to the VictoryCharts equivalent: https://formidable.com/open-source/victory/docs/victory-bar/

The purpose of having our components wrap VictoryCharts is to prescribe stylistic choices, such as colors, line weight, spacing, etc.

**Todo:**
- [x] Add VictoryCharts as a dependency
- [x] Rewrite the LineChart component
- [x] Add unit tests
- [x] Update the LineChart story

The basic line chart should look like this, and have all of the components shown in this chart.
<img width="515" alt="lc no tooltip" src="https://user-images.githubusercontent.com/7065695/39656140-ad36b3fe-4fb2-11e8-935a-336486b28b65.png">
Shown with common tooltip component:
<img width="514" alt="lc tooltip" src="https://user-images.githubusercontent.com/7065695/39656139-ad17ee88-4fb2-11e8-8a1f-0297b265ad2f.png">

Generally, I'm expecting that the color and font styling will be inherited from a common VictoryTheme, so don't worry about the color and font for this issue.

In addition to what is shown, BarChart should have the following features:
- [ ] Accessibility description
- [x] Responsive

jaronheard:
I've updated the design of the line chart to the latest version that shows with and without the common tooltip component.

AlexanderNull:
Accessibility tracked in https://github.com/hackoregon/civic/issues/122

PR #95, opened 2018-05-06
First iteration of Victory line chart

cp:
Not sure how complete I should make this, so totally open to feedback.

Questions that came up while building this:
- Should we define some shared data interface? How would it work with multi-dimensional data?
- Should we introduce a ChartContainer component that renders titles and subtitles?
- How should we handle formatting data?

https://github.com/hackoregon/civic/issues/48


package.json
```diff
@@ -110,6 +110,8 @@
     "webpack-node-externals": "^1.5.4"
   },
   "dependencies": {
-    "@hackoregon/civic-server": "0.0.10"
+    "@hackoregon/civic-server": "0.0.10",
+    "victory": "^0.26.0",
+    "webpack-blocks": "^1.0.0"
```
AlexanderNull: Are these only used in the components-library package? If so can these get moved down to that local package.json?

packages/component-library/src/LineChart/LineChart.js
```diff
@@ -0,0 +1,89 @@
...
+        />
+        <VictoryLine
+          data={data.map(d => ({ dataKey: d[dataKey], dataValue: d[dataValue] }))}
+          x="dataKey"
+          y="dataValue"
+        />
+        <VictoryScatter
+          data={data.map(d => ({ dataKey: d[dataKey], dataValue: d[dataValue] }))}
+          x="dataKey"
+          y="dataValue"
```
AlexanderNull: If we're explicitly defining the data shape via the map functions on 61 and 66 then we can just define the keys as x/y and then omit these additional properties as the defaults for x and y are just "x" and "y". If we're going to stick with "dataKey" and "dataValue" then we should use a constant for the object keys and the props here so that anyone coming here later understands the coupling of these lines.

packages/component-library/src/LineChart/LineChart.js
```diff
@@ -0,0 +1,89 @@
...
+          tickFormat={data.map(d => d[dataKeyLabel])}
+        />
+        <VictoryAxis
+          dependentAxis
+          tickValues={data.map(d => d[dataValue])}
+        />
+        <VictoryLine
+          data={data.map(d => ({ dataKey: d[dataKey], dataValue: d[dataValue] }))}
+          x="dataKey"
+          y="dataValue"
```
AlexanderNull: See below comment

packages/component-library/src/LineChart/LineChart.js
cp: Hey, we talked a bit about this issue during the meeting, and @DingoEatingFuzz posted some notes in Slack. If it's okay with you, I say we keep this for now (it's consistent with the Bar chart) then see how we can possibly standardize on a better common format as we continue to merge these charts in.

package.json
cp: Good call. Pushed up a fix, thanks!

AlexanderNull:
::shipit::

Issue #67, opened 2018-05-01
Add a platform-wide number formatting utility to the component library


jaronheard:
A common number formatting utility needs to written and integrated into the component library.  It should take a number and format it to a set of platform wide-specifications. It should support multiple platform-wide presets. This number formatter should be a more restrictive interface that just uses d3-format underneath it.

- [x] Determine rules for a few number presets
- [x] Write the number-formats utility - see comment below for architectural details
- [x] Add unit tests

**Original context below:**
I think we should create a number formatter to ensure consistent display of numbers across our platform. Originally, I was thinking this would be useful for data visualizations, but I think that it would be useful across the platform. I think that this should be a component or something that lives on the platform and can be updated.

It should take a number and format it to a set of platform wide-specifications. It should support multiple platform-wide presets.

There should be some way to indicate the precision - or number of significant digits.

I'm not sure on the technological architectural details, so if someone has an idea on how this should work, please chime in. There may be a tool that we can use with common presets across our projects.

For a basic data visualization preset, use the following guidelines:
* Commas
    * Always use commas for thousands separator
* Numbers and decimal places
    * Use units that will result in showing a maximum of 6 digits
        * Fewer digits is better
    * Lean towards showing 0-1 decimal places. 
        * Should reflect the precision of the data
        * Show whole dollars or dollars and cents (e.g. $123 or $1.23)
    * Examples
        * $7.3 million, not $7,300,000
        * 24%, not 23.79%
        * $535,000
* Show units (e.g. dollar sign, percent)

Thoughts? @DingoEatingFuzz

DingoEatingFuzz:
I think this is a good idea. Already present in some places is the use of [`d3-format`](https://github.com/d3/d3-format) to do things like commas and significant digits.

It might be nice to create a `utils/number-formats` that can be used as a more restrictive interface that just uses `d3-format` underneath it.

e.g.,

```js
import { number, percent } from '@hackoregon/component-library/utils/number-formats';

let count = 7300000;
let countText = number(count); // 7.3 million

let pct = 0.2379;
let pctText = percent(pct); // 24%
```

With the implementation of `number-formats` looking something like

```js
import { format } from 'd3-format';
export const percent = d3.format(".0%"); // returns a function that takes a number and returns a string
```

jaronheard:
#96 addressed this.

PR #96, opened 2018-05-06
Formatter utilities

cp:
I believe this should cover all the cases described in the issue. Should be pretty trivial to add more as we go on.

https://github.com/hackoregon/civic/issues/67


packages/component-library/src/utils/formatters.js
```diff
@@ -0,0 +1,18 @@
+import { format } from 'd3-format';
+
+export const numeric = (d) => {
+  let formatted;
+
+  if (d >= 1000000) {
+    formatted = format('.2s')(d)
+      .replace('M', ' million')
```
cp: Any ideas on how I could make this less rote?

DingoEatingFuzz: A couple other caveats here: this won't work for billions, which uses `G` since it's scientific notation. It also won't work for trillions (which we could run into if we do budget projects in the future).

Given all these caveats, it may make more sense to write our own formatter rather than hacking on top of d3's formatter in this case.

packages/component-library/src/utils/formatters.js
```diff
@@ -0,0 +1,18 @@
...
+      .replace('.0', '');
+  } else {
+    formatted = format(',.0f')(d);
+  }
+
+  return formatted;
+};
+
+export const percentage = format('.0%');
+export const dollars = d => `$${d}`;
```
DingoEatingFuzz: Pretty nice how these compose.

packages/component-library/src/utils/formatters.test.js
```diff
@@ -0,0 +1,20 @@
+import { numeric, percentage, dollars } from './formatters';
+
+describe('formatters', () => {
+  it('should format numbers correctly', () => {
+    expect(numeric(0.1)).to.eql('0');
+    expect(numeric(0.5)).to.eql('1');
+    expect(numeric(1000)).to.eql('1,000');
+    expect(numeric(1000000)).to.eql('1 million');
+    expect(numeric(7300000)).to.eql('7.3 million');
+  });
```
DingoEatingFuzz: Might make sense to add tests for billions and trillions for the caveats mentioned above.

packages/component-library/src/utils/formatters.test.js
```diff
@@ -0,0 +1,20 @@
...
+
+  it('should format percentages correctly', () => {
+    expect(percentage(0.75)).to.eql('75%');
+    expect(percentage(0.2379)).to.eql('24%');
+  });
+
+  it('should format dollars correctly', () => {
+    expect(dollars(12)).to.eql('$12');
```
DingoEatingFuzz: Could be nice to write tests for larger dollars, fractions of dollars, and dollars working with numeric. IMO tests are a great way to also document usage.

cp:
@DingoEatingFuzz @jaronheard Okay, I've implemented better support for millions, billions, trillions, and expanded on the tests. I think we can continue to build on top of this, but I feel pretty good about it.

Let me know if you think it makes sense to split into two formatters, but I think this may be the best for now from a usability prospective.

jaronheard:
@DingoEatingFuzz I think if you approve @cp's changes this is ready to be merged.

Issue #83, opened 2018-05-04
Create HorizontalBarChart component


jaronheard:
A new HorizontalBarChart component needs to be written using VictoryCharts:
https://formidable.com/open-source/victory/docs/victory-bar/

The purpose of having our components wrap VictoryCharts is to prescribe stylistic choices, such as colors, line weight, spacing, etc.

**Todo:**
- [x] Add VictoryCharts as a dependency
- [x] Write HorizontalBarChart component
- [x] Add unit tests
- [x] Update the HorizontalBarChart story

The basic horizontal bar chart should look like this, and have all of the components shown in this chart.
<img width="308" alt="hb no toolkit" src="https://user-images.githubusercontent.com/7065695/39656425-87f04446-4fb4-11e8-8923-ba8b060845e0.png">
Shown with common tooltip component:
<img width="302" alt="hb toolkit" src="https://user-images.githubusercontent.com/7065695/39656429-8ce1d7c6-4fb4-11e8-82d2-34067b596ecb.png">

Generally, I'm expecting that the color and font styling will be inherited from a common VictoryTheme, so don't worry about the color and font for this issue.

In addition to what is shown, HorizontalBarChart should have the following features:
- [x] Accessibility description
- [x] Responsive

AlexanderNull:
Adding in unit tests

jaronheard:
#267 and others addressed

PR #107, opened 2018-05-06
Make the horizontal bar chart horizontal, and make new bar chart component

accua:
This PR addresses #83. It makes the horizontal bar chart, horizontal and creates a new bar chart component. This also sets up testing files for both but doesn't actually test. I don't have much experience testing JS yet, but wanted to get this up for review.

~~* Still needs tooltip, and bar highlighting on hover.~~
* Still needs tooltip font color changed to `tomato`, and tooltip-border == none


DingoEatingFuzz:
This looks good in that it correctly makes the HorizontalBarChart horizontal. However it introduces a new issue: we have no normal bar chart.

Do you want to extend this PR to handle both chart types in some graceful DRY way, or should I make another issue?

accua:
@DingoEatingFuzz forgot to push up the other chart, this addresses that and adds a tooltip/some styling.

AlexanderNull:
I thought you had copied over some tests for this

accua:
@AlexanderNull not yet. I heard that #95 was a good place to look for a start.

@jaronheard I'm not sure, but I would imagine so. I'll do some research! I will begin working on a tooltip component unless someone else already has?

Issue #62, opened 2018-04-26
Build react-map-gl component


jaronheard:
We are moving from leaflet to using mapbox, react-map-gl and deck.gl

This map will be the basic map for our visualization to be built on.
MORE DETAILS ON MAP TILES, DEFAULTS, ETC

Todo:
- [ ] Add necessary dependencies
- [ ] Write the ReactMap component
- [ ] Add unit tests
- [ ] Update the ReactMap story

jaronheard:
Building a basic map component with these technologies can be started on.

znmeb:
Is there a guide on what MapBox needs from APIs and / or databases?

DingoEatingFuzz:
mapbox-gl-js uses either a vector tile server or a geojson object. I suspect we will only be using geojson.

Read more here: https://www.mapbox.com/mapbox-gl-js/api/

znmeb:
Thanks!!

jaronheard:
#114 addresses!

PR #114, opened 2018-05-07
Initial iteration of the React Map component

mendozaline:
This is the basic map component for issue [#62](https://github.com/hackoregon/civic/issues/62). I added react-map-gl, created a new React Map component and story for it. 

I want to get this reviewed before moving on to the deck.gl overlays.

A testing file still needs to be added and written. I'll need help with that.

Some questions that we still need to resolve include what to do about a default map style and any other defaults we want to set.


packages/component-library/src/ReactMap/ReactMap.js
```diff
@@ -0,0 +1,60 @@
+import React, { PropTypes } from 'react';
+import MapGL from 'react-map-gl';
+import './mapbox-gl.css';
+
+const ReactMap = (props) => {
```
cp: Might make sense to simplify the name of this component (and file) to just `Map`. Thoughts?

mendozaline: That makes sense. I was trying to differentiate it from the Leaflet Map component already there. I'll rename it.

mendozaline:
I renamed the component from ReactMap to BaseMap and converted it from a functional to class component. This allows the map to now pan, zoom, and resize.

I also included and wrote tests for the component.

packages/component-library/src/BaseMap/BaseMap.js
```diff
@@ -0,0 +1,82 @@
...
+  margin: auto;
+  max-width: 900px;
+`;
+
+export default class BaseMap extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      viewport:{
+        width: window.innerWidth > 900 ? 898 : window.innerWidth,
```
jaronheard: May change this to work with responsive containers if it doesn't already

mendozaline: I was meaning to change that. It's already inside a div and it's responsive, so the map resize when the window changes and stays centered. But I'll update it.

PR #129, opened 2018-05-11
Rename GeoJSON layer component to MapOverlay and keep previous BaseMap Component

jasonleonhard:
1. Add deck.gl GeoJSON layer components to layer on BaseMap
2. Rename GeoJSON layer component to MapOverlay and keep previous BaseMap Component


packages/component-library/src/MapOverlay/MapOverlay.js
```diff
@@ -0,0 +1,99 @@
...
+
+const colorScale = r => [r * 255, 140, 200 * (1 - r)];
+
+// Source data GeoJSON (currently only Vancouver BC)
+const DATA_URL =
+  'https://raw.githubusercontent.com/uber-common/deck.gl-data/master/examples/geojson/vancouver-blocks.json'; // eslint-disable-line
+
+export default class BaseMap extends Component {
+  constructor(props) {
+    super(props);
```
AlexanderNull: Fun trivia fact: super(props) isn't needed unless you're trying to call `this.props` within the constructor or you're calling a separate class method within the constructor that also tries to reference `this.props`

packages/component-library/src/MapOverlay/MapOverlay.js
```diff
@@ -0,0 +1,99 @@
...
+};
+
+BaseMap.propTypes = {
+  mapboxStyle: PropTypes.string,
+  mapboxToken: PropTypes.string.isRequired,
+};
+
+BaseMap.defaultProps = {
+  mapboxStyle: "mapbox://styles/themendozaline/cjg6296ub04ot2sqv9izku3qq",
+};
```
AlexanderNull: There's a lot of duplicated code between this and the other BaseMap component. Would be good to see the existing BaseMap extended if needed.

packages/component-library/src/MapOverlay/MapOverlay.js
```diff
@@ -0,0 +1,99 @@
...
+        maxZoom: 20,
+        pitch: 0,
+        bearing: 0,
+      },
+      data: null
+    };
+
+    fetch(DATA_URL)
+      .then(resp => resp.json())
+      .then(data => this.setState({data}));
```
AlexanderNull: Should this be done up in a store?

PR #131, opened 2018-05-15
Create PullQuote component and story

lowesusannah:
Here is a first attempt at the PullQuote component and story. Thanks!


packages/component-library/src/index.js
```diff
@@ -25,5 +28,9 @@ export { default as Logo } from './Logo/Logo';
 export { default as ShareCollection } from './Share/ShareCollection';
 export { default as isClient } from './utils/isClient';
 export { default as CivicVictoryTheme } from "./VictoryTheme/VictoryThemeIndex";
+export { default as CollectionHero } from './Hero/CollectionHero'
+export { default as BaseMap } from './BaseMap/BaseMap';
+export { default as PullQuote } from './PullQuote/PullQuote';
```
cp: Looks like we'll either need to move `PullQuote` into its own directory, or update our import here to import it `from './PullQuote'`. We might want to do the former to be inline with the rest of our components.

jaronheard:
@cp @lowesusannah I resolved a merge conflict and relocated the PullQuote.js file into it's own directory. 

The next step would be to add tests. I'm ok with either merging this and adding a new issue to add tests or adding in tests here before merging.

Issue #147, opened 2018-05-18
Add unit tests for ScatterPlotMap


mendozaline:
This is a follow up on #134 , which still needs unit tests.

- [x] Add unit tests

PR #160, opened 2018-05-22
Scatterplot Map Unit Tests

mendozaline:
This addresses [#147](https://github.com/hackoregon/civic/issues/147). I refactored the ScatterPlot Map layer and added unit tests for the component.


packages/component-library/src/BaseMap/BaseMap.js
```diff
@@ -21,12 +21,12 @@ export default class BaseMap extends Component {
       viewport:{
```
jaronheard: Nice improvements

PR #165, opened 2018-05-22
Resolve upstream with fork

jasonleonhard:
An assortment of commits of varying types. I have confirmed my build works and is running node 8.11.1  and all should be ready.


DingoEatingFuzz:
Scrolling through the diff everything looks good, but I can't accept this PR the way it is. Namely because of the name and lack of detailed description.

Instead, can you break this into a PR for each logical grouping of changes? I see two:

1. Using the `prop-types` lib instead of React
2. Adding tests to the HexOverlay component

Once you do that, you can close this PR and get reviews on each new PR.

Thanks, and sorry to be a stickler on PR formats, I just don't want to cultivate the habit of working on a bunch of different things and then "syncing".

jasonleonhard:
Will do a new request.

PR #187, opened 2018-05-30
Scale icons based on zoom level on IconMap

mendozaline:
Currently the icons on the IconMap stay the same size when you zoom in or out. This changes that so icons get smaller when you zoom out and bigger as you zoom in. It also adds visibility props. And I updated the unit tests.


mendozaline:
This latest change should address the issue you brought up about icon size at closer zoom levels.

PR #198, opened 2018-06-03
Remove filson and others

### LINKED ISSUE #175: Remove Typekit and replace all instances of Filson with Rubik
Typekit is currently used to download Filson, which is not a free font. It works locally, but our domain isn't whitelisted so it doesn't work once published. 

**Working vs. Not Working**
image
image

Since we are already also using Rubik for headings in places, the path forward is to replace all instances of Filson with Rubik and remove our dependency on Typekit.

DingoEatingFuzz:
Typefaces on the chopping block:

  1. FilsonSoft
  2. Lato
  3. OpenSans

FilsonSoft is what we were using for titles, namely StoryCard titles. It has to go due to it requiring a TypeKit account.

Lato was only used in `.panel` elements and OpenSans was never really used. They are being removed because we don't need them. Merriweather is our body typeface and Rubik is our titles typeface.

This brings the font count down from six to three.

Closes #175

### DIFF
--- a/packages/2017/src/components/RootPage.js
+++ b/packages/2017/src/components/RootPage.js
@@ -1,13 +1,11 @@
 import React from 'react';
 import { connect } from 'react-redux';
-import Typekit from 'react-typekit';
 import { Header } from '@hackoregon/component-library';
 
 const isRoot = ({ pathname }) => pathname === '/';
 
 export const RootPage = props => (
   <div>
-    <Typekit kitId="mbf2sam" />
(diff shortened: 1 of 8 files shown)


jaronheard:
I think it says “have we hit peak farmers market” in two spots - the Filson
is the one right above the graph further down the page.

DingoEatingFuzz:
Right you are. I found it and removed it from ChartContainer. Now those elements will get their font family from the global stylesheet.

PR #205, opened 2018-06-06
Page Layout and General Style Updates

danieltomasku:
This PR improves the layout styling of the `/farmers-markets` collection example and other global components like the header and footer. This gives us a further start to implementing collections for the all the project teams. After this PR is merged in we can fine tune the design.

Updates to the landing page are coming soon. Stay tuned 😃


packages/component-library/src/Navigation/Nav.js
```diff
@@ -10,16 +10,17 @@ const defaultMenu = [
...
+      { name: 'Neighborhood Development', path: '/neighborhood' },
+      { name: 'Transportation Systems', path: '/transportation' },
+      { name: 'EXAMPLE Farmers Markets', path: '/farmers-markets' },
     ],
   },
   {
-    name: 'About',
-    path: '/about',
+    name: 'Sandbox',
+    path: '/sandbox',
```
DingoEatingFuzz: Oh shoot, I just now noticed that these changes are in the component library. By implementing the new menu in this way, you inadvertently broke the 2017 menu.

It's weird, but the "default" menu is the 2017 menu, which I left untouched. Instead, I overrode the menu for the 2018 package in RootPage for the header you removed.

https://github.com/hackoregon/civic/blob/bfcd3d570f2a048b6e75900b906ccfc5795c184f/packages/2018/src/components/RootPage.js#L5-L22

Since the header for the 2018 site is now being added to the App component for each template, I think the easiest path forward is to keep this default menu as the 2018 menu and instead use the pattern I was using for 2018 in the 2017 package.

danieltomasku:
@DingoEatingFuzz just added the menu items back to the 2017 package. i would love to get this merged in and then iterate on it later!

jaronheard:
Follow-up issue: #221

PR #226, opened 2018-06-07
Remove duplicate imports of stories

jaronheard:
`mapOverlayStory` and `hexOverlayStory` were imported twice, causing an error when running yarn storybook.

This PR fixes that error.


jaronheard:
Added one more line change fix error on storybook run

Issue #213, opened 2018-06-07
Civic Sandbox Map Component


mendozaline:
This map component should be a composite layer of other primitive layers like polygon, scatterplot, path, etc. It will be used to show foundation, slide and boundary data.

To do:
- [x] Add the CivicSandboxMap component
- [x] Add CivicSandboxMap story
- [x] Add tooltips

PR #229, opened 2018-06-08
Civic Sandbox Map Component

mendozaline:
This addresses issue [#213](https://github.com/hackoregon/civic/issues/213).

I created a new CivicSandbox Map component which allows multiple layers to be drawn on the same map canvas. I added a CivicSandbox Map story. I also created a CivicSandbox Map tooltip.

This is my first iteration and there's much room for improvement in terms of color choice, design, etc.


AlexanderNull:
There was a small merge conflict so I pushed a fix for that back up. Merge in if it looks good.

PR #230, opened 2018-06-08
Add data table component

dsdemaria:
I thought I should PR up what I have so far. I can add more as time goes on (like tests hehe) but I wanted to get this out there.

I added some basic instructions on how to use the component in the storybook.


dsdemaria:
I'd like to do more with it. Things like:
- actually write tests
- improve documentation
   * write out more storybook examples
   * flesh out the how to use documentation
- split up large file into duck-like folder
- document lessens learned for making more components reusable
- build in themeing

jaronheard:
@dsdemaria That sounds great! However, I think that merging in this will give us the right base to build on - we can continue to iterate on the component, but I think that it's unlikely we will have to change the architecture in a way that is breaking if we start making things in projects with this now.

I had the same feeling about the things that I want to do for our data visualization components, but now's the time to get things merged and start building with that we've got!

Issue #219, opened 2018-06-07
Clean up and standardize props for chart components


jaronheard:
Our data visualization chart components have some props that should be somewhat standardized across our components in order to be most useful:

- [x] Currently, our pattern is to use dataKey and dataValue in order to allow our components to use data that is structured properly but does not need to conform to a naming conventions. This convention should be extended to categories.

- [x] Synchronize scatterplot set default domain prop logic across the other components

- [x] Add a prop to format the X and Y axis - mostly this will be taking in the formatters from our standard formatters (https://github.com/hackoregon/civic/blob/master/packages/component-library/src/utils/formatters.js), but could take in any function.

jaronheard:
WIth #266, this issue should be complete

PR #232, opened 2018-06-09
Set updated pattern for charts in Scatterplots

jaronheard:
This PR addresses #219, items 1 and 3.

Next step is to synchronize the scatterplot set default domain prop logic across the other components

I thought I'd get this code out there for review as I work to standardize it to the other chart components.


packages/component-library/src/Scatterplot/Scatterplot.js
```diff
@@ -186,6 +84,7 @@ const Scatterplot = ({
         <VictoryAxis
           animate={{ onEnter: { duration: 500 } }}
           style={{ grid: { stroke: 'none' } }}
+          tickFormat={x => xNumberFormatter(x)}
```
DingoEatingFuzz: you can instead just pass the formatter in directly: `tickFormat={xNumberFormatter}`.

packages/component-library/src/Scatterplot/Scatterplot.js
```diff
@@ -202,6 +101,7 @@ const Scatterplot = ({
                   : 'none',
             },
           }}
+          tickFormat={y => yNumberFormatter(y)}
```
DingoEatingFuzz: Same thing here.

packages/component-library/src/Scatterplot/Scatterplot.js
```diff
@@ -228,12 +128,12 @@ const Scatterplot = ({
         </VictoryPortal>
         <VictoryScatter
           animate={{ onEnter: { duration: 500 } }}
-          categories={{ x: dataKeyLabel }}
+//        categories={{ x: categoryData }}
```
DingoEatingFuzz: No more categories?

jaronheard: `categories` actually doesn't do anything for the VictoryScatter - it actually gives you an invalid prop warning when you pass it in - but it is a common prop across the other chart components.

jaronheard: Have it commented out here because for the other components we will want it.

packages/component-library/src/Scatterplot/Scatterplot.js
jaronheard: Sweet, will do 👍

jaronheard:
Additional commits to address item 2 of #232 - think this will set us up pretty well to make our data visualizations without having to make many breaking changes.

There's kind of a lot in here, and definitely still work to be done on the display.

packages/component-library/src/HorizontalBarChart/HorizontalBarChart.js
```diff
@@ -11,64 +11,52 @@ import {
...
     fontSize: '14px',
     fontWeight: 'bold',
   };
 
+  const barData =
+    sortOrder && sortOrder.length
+      ? data
+      : data.map((d, index) => {
+        const e = d;
+        e.defaultSort = index + 1;
```
DingoEatingFuzz: `return { ...d, defaultSort: index + 1 }`

PR #234, opened 2018-06-10
New working with redux guide

DingoEatingFuzz:
[Link to formatted file](https://github.com/hackoregon/civic/blob/redux-guide/WORKING_WITH_REDUX.md).

### DIFF
--- a/WORKING_WITH_REDUX.md
+++ b/WORKING_WITH_REDUX.md
@@ -1,0 +1,293 @@
+# Working with Redux
+
+Redux is how state flows through Civic applications. It is an implementation of the [Flux Architecture](https://facebook.github.io/flux/) that's used to handle user and API interactions ([actions](https://redux.js.org/basics/actions)), manage state transformations ([reducers](https://redux.js.org/basics/reducers)), and provide data to UI components ([selectors](https://redux.js.org/recipes/computing-derived-data)). It does this by maintaining a single global, immutable store.
+
+
+
+## Brief Glossary
+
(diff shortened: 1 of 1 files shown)


DingoEatingFuzz:
> It wasn’t obvious to me that you needed to add the reducer for each component to the index.js file. Also the error message for this was not helpful, so might be worth adding a few words.

Yes! I definitely need to add that.

> I think an example that involved importing and doing data transformations would be more useful and realistic.

I'm not following this one. Selectors don't import anything and the example does do data transformation.

Do you have an example in mind for what the example should be?

jaronheard:
Like this but done properly? :P

https://github.com/hackoregon/civic/blob/53bd7cae6b3c4b9f9f45ab6a0cfe8f2e7a2db0b0/packages/2018-transportation-systems/src/state/decline-in-ridership/selectors.js

In the example below you have the getFarmersMarketsOverTimeData selector. I
wanted to see how a data transformation would interact with that selector
(e.g. should it be part of the same selector? should it be a separate
selector and composed somehow?). I think if you just extended your current
example to show that it would be fantastic.
https://github.com/hackoregon/civic/blob/master/packages/2018-example-farmers-markets/src/state/farmers-markets-over-time/selectors.js

DingoEatingFuzz:
@jaronheard I beefed up the selector examples and added a note about how reducers from ducks need to be included in the project reducer. Lemme know what you think!

jaronheard:
@DingoEatingFuzz That looks great! Going to go ahead and merge this in.

PR #244, opened 2018-06-12
adds package selector component and story

nicrocs:
This component will be used for to select a package for the civic sandbox.


jaronheard:
If someone wants to review this, it shouldn't be breaking anything now

PR #246, opened 2018-06-12
Card - plateau in ridership

jaronheard:
It's a card!

I also added the redux utilities for the transportation project: import-adapter, import-adapter-reducer, import-adapter-action-emitter - this will need to be done once for each of the teams.

This is all directly based on the pattern in the FarmersMarketsOverTime card.


packages/2018-transportation-systems/src/components/DeclineInRidership/index.js
```diff
@@ -1,25 +1,93 @@
...
+  isRidershipOverTimePending,
+  getRidershipOverTimeData,
+} from '../../state/decline-in-ridership/selectors';
+
+const sampleSimpleData = [
+  { x: 0, y: 20 },
+  { x: 10, y: 30 },
+  { x: 20, y: 50 },
+  { x: 30, y: 40 },
+];
```
DingoEatingFuzz: Looks like this sample data is no longer being used.

packages/2018-transportation-systems/src/components/DeclineInRidership/index.js
```diff
@@ -1,25 +1,93 @@
...
+  background: #EEE;
+`;
+
+const cardError = css`
+  width: 100%;
+  padding: 50px;
+  text-align: center;
+  background: #FDD;
+  border: 1px solid #C99;
+`;
```
DingoEatingFuzz: Not gonna make you do it as part of this PR, but these common loading and error states should both be moved into the component library and redone to look nice.

packages/2018-transportation-systems/src/components/DeclineInRidership/index.js
```diff
@@ -1,25 +1,93 @@
...
+      return <div className={cardLoading}>Loading...</div>;
+    } else if (!ridershipOverTime) {
+      return <div className={cardError}>Could not render Plateau in Ridership</div>;
+    }
+
     return (
       <CivicStoryCard
-        title="Decline in Ridership"
+        title="Plateau in Ridership"
         slug="decline-in-ridership"
```
DingoEatingFuzz: We'll probably want to update the slug for this code too. Which needs to be done here and in `registry.js` in the same package.

packages/2018-transportation-systems/src/state/decline-in-ridership/actions.js
```diff
@@ -0,0 +1,22 @@
...
+
+export const IMPORT_START = 'DECLINE_IN_RIDERSHIP_OVER_TIME/START';
+export const IMPORT_SUCCESS = 'DECLINE_IN_RIDERSHIP_OVER_TIME/SUCCESS';
+
+export const ridershipOverTimeStart = actionEmitter(IMPORT_START);
+export const ridershipOverTimeSuccess = actionEmitter(IMPORT_SUCCESS);
+
+const DECLINE_IN_RIDERSHIP_API = 'http://service.civicpdx.org/transportation-systems/passenger-census/system/annual/total/?format=json';
+
+const importPromise = axios.get(DECLINE_IN_RIDERSHIP_API);
```
DingoEatingFuzz: It's great that the `importAdapter` still works here since it just takes a promise, but I think it would be better to make an `apiAdapter` that is just like the `importAdapter` but takes a URL.

This way `axios` becomes an implementation detail that not every duck has to be concerned with.

packages/2018-transportation-systems/src/state/decline-in-ridership/actions.js
```diff
@@ -0,0 +1,22 @@
...
+
+export const fetchRidershipOverTime = importAdapter(
+  importPromise,
```
DingoEatingFuzz: So this argument would become `DECLINE_IN_RIDERSHIP_API`.

packages/2018-transportation-systems/src/state/decline-in-ridership/selectors.js
```diff
@@ -0,0 +1,37 @@
...
+    year: yearObj.year,
+    ons: yearObj.saturday_sum_ons / 52,
+  }
+  ))).concat(data.data.map(yearObj => ({
+    type: 'Sunday',
+    year: yearObj.year,
+    ons: yearObj.sunday_sum_ons / 52,
+  }
+  )))
+);
```
DingoEatingFuzz: This is a really cool selector. Can you add a test for it?

packages/2018-transportation-systems/src/state/decline-in-ridership/selectors.js
```diff
@@ -0,0 +1,37 @@
...
+
+export const isRidershipOverTimePending = createSelector(
+  getRidershipOverTimeRequest,
+  ({ pending }) => !!pending,
+);
+
+const calculateChangeSince2002 = data => {
+  let ret;
+
+}
```
DingoEatingFuzz: I can't imagine this is being used. Is it safe to delete?

packages/2018-transportation-systems/src/state/decline-in-ridership/actions.js
```diff
@@ -0,0 +1,22 @@
+import axios from 'axios';
+
+import importAdapter from '../import-adapter';
+import actionEmitter from '../import-adapter-action-emitter';
+
+export const IMPORT_START = 'DECLINE_IN_RIDERSHIP_OVER_TIME/START';
+export const IMPORT_SUCCESS = 'DECLINE_IN_RIDERSHIP_OVER_TIME/SUCCESS';
```
DingoEatingFuzz: Since this is dealing with an API rather than a file, it's really important we handle the `ERROR` action. 

With the way this is written now, if the API request fails (i.e., responds with a 4xx or 5xx status code), the card will stay in the loading state and the user will have no indication that the request is no longer loading, it's just broken.

packages/2018-transportation-systems/src/state/import-adapter.js
```diff
@@ -0,0 +1,11 @@
+// A redux action utility for making a thunk based on an async file import
+// ex. const someJsonInRepo = importAdapter('../assets/some.json', { start, success });
+//     dispatch(someJsonInRepo())
+const importAdapter = (importPromise, { start, success }) => () => (dispatch) => {
+  dispatch(start());
+  return importPromise.then((res) => {
```
DingoEatingFuzz: Continuing the thought above about making this an `apiAdapter`, this line would become something like:

```js
return axios.get(url).then((res) => {
  dispatch(success(res));
}).catch((err) => {
  dispatch(failure(err));
});
```

packages/2018-transportation-systems/src/components/DeclineInRidership/index.js
jaronheard: Agreed. Created issue #249.

packages/2018-transportation-systems/src/components/DeclineInRidership/index.js
jaronheard: Yep, removed.

packages/2018-transportation-systems/src/components/DeclineInRidership/index.js
jaronheard: 👍

packages/2018-transportation-systems/src/state/decline-in-ridership/selectors.js
jaronheard: removed

jaronheard:
Got something working for the API reducer but I'm having trouble figuring out why these tests are failing in the latest code that I pushed

```
2) ridership-over-time ridership-over-time reducer should handle API_ERROR:

      AssertionError: expected { pending: false, error: undefined } to deeply equal { pending: false, error: 'error' }
      + expected - actual

       {
      -  "error": [undefined]
      +  "error": "error"
         "pending": false
       }
      
      at Context.<anonymous> (src/state/decline-in-ridership/decline-in-ridership.test.js:108:14)

  3) ridership-over-time ridership-over-time selectors getRidershipOverTimeRequest extends the root selector:
     AssertionError: expected undefined to deeply equal { one: 'two', three: 4 }
      at Context.<anonymous> (src/state/decline-in-ridership/decline-in-ridership.test.js:129:16)
```

packages/2018-transportation-systems/src/state/decline-in-ridership/decline-in-ridership.test.js
```diff
@@ -0,0 +1,215 @@
...
+          ons: 217309.5,
+        }, {
+          type: 'Saturday',
+          year: 2001,
+          ons: 73028,
+        }, {
+          type: 'Sunday',
+          year: 2001,
+          ons: 71360.5,
+        }];
```
DingoEatingFuzz: 👏 👏 Seeing a concrete usage of a selector helps so much when figuring out what it does.

packages/2018-transportation-systems/src/state/decline-in-ridership/selectors.js
jaronheard: Added test

jaronheard:
@DingoEatingFuzz I tried to write tests for the API adapter, but I’m afraid it’s not the task for me...

This could be relevant for the next person who attempts it. https://github.com/node-nock/nock/issues/699

PR #255, opened 2018-06-14
HexOverlay Tooltip implimented

jasonleonhard:
Only issue is the boolean knobs in HexOverlay, would like to get another set of eyes on that one minor issue, for now I see no reason not to merge this, it is ready. Thanks.


DingoEatingFuzz:
You have an 80k+ LOC json file in this PR. We can't merge that.

packages/component-library/src/MapTooltip/MapTooltip.js
```diff
@@ -24,6 +24,7 @@ const MapTooltip = (props) => {
     primaryField,
     secondaryName,
     secondaryField,
+    isHex
```
DingoEatingFuzz: Ideally we don't have component-specific switches like this in general components. If the hex overlay needs its own code path, that is good evidence that the abstractions aren't right.

Either hex overlay needs its own tooltip or maptooltip's api needs to expand in a way to support more use cases.

@dorenproctor and I had a conversation about this on Monday.

packages/component-library/stories/HexOverlay.story.js
```diff
@@ -85,17 +92,55 @@ const demoMap = () => {
...
+        opacity={opacity}
+        coverage={coverage}
+        radius={radius}
+        elevation={elevation}
+        colorRange={colorRange}
+        lightSettings={lightSettings}
+        filled={filled}
+        wireframe={wireframe}
+      >
+        <MapTooltip isHex={true} />
```
DingoEatingFuzz: Requiring the user of HexOverlay to know that the tooltip needs this `isHex` flag to not throw errors is also a really bad developer experience.

jasonleonhard:
Implemented d3 fetch to avoid large json file in git.

jasonleonhard:
Currently MapTooltip does require isHex to be true like you pointed out:

    <MapTooltip isHex={true} />

We can discuss alternatives tomorrow and/or we can work on new StoryCards like we discussed and consider this later. Either way is fine with me, let's discuss tomorrow. Thanks.

packages/component-library/src/MapTooltip/MapTooltip.js
jasonleonhard: Currently MapTooltip does require isHex to be true like you pointed out:

    <MapTooltip isHex={true} />

We can discuss alternatives tomorrow and/or we can work on new StoryCards like we discussed and consider this later. Either way is fine with me, let's discuss tomorrow. Thanks.

PR #257, opened 2018-06-14
MapOverlay Elevation and ToolTip

jasonleonhard:
Now does not need user to click extruded each time to change the elevation, it works as it should.


jasonleonhard:
implimented d3 fetch of data in lieu of a local json file that was very very large.

DingoEatingFuzz:
@jasonleonhard, can you address these merge conflicts, please.

Issue #258, opened 2018-06-14
Get geojson out of the component library


DingoEatingFuzz:
Four map components are megabytes in size. This is most definitely contributing to the size and build time of storybook

![image](https://user-images.githubusercontent.com/174740/41389762-6d806aa2-6f47-11e8-82e9-da637e61e56f.png)

This amounts to a src of nearly 20Mb.

![image](https://user-images.githubusercontent.com/174740/41389807-b087485c-6f47-11e8-97b5-29a587b4bd32.png)

Fortunately these geojson files don't impact the project and year packages, but they are still impacting the utility of storybook.

PR #264, opened 2018-06-15
Loading GeoJSON data via d3

mendozaline:
This address issue [#258](https://github.com/hackoregon/civic/issues/258).

I deleted the GeoJSON file from the ScreenGridMap component. Now that data is loaded via d3.


packages/component-library/stories/ScreenGridMap.story.js
```diff
@@ -1,83 +1,99 @@
...
+    super(props);
+    this.state = {
+      data: null,
+      error: null,
+    };
+  }
+
+  componentDidMount() {
+    let cmp = this;
+    d3.json(this.props.url, (error, data) => {
```
DingoEatingFuzz: You could use [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Body/json) here to avoid bringing in d3.

packages/component-library/stories/ScreenGridMap.story.js
```diff
@@ -1,83 +1,99 @@
...
+      if (error) { return this.setState({error: error})}
+      cmp.setState({data: data});
```
DingoEatingFuzz: There's no need to store the component in a variable outside this callback since you're using an arrow function. `this.setState({ data });` works.

packages/component-library/stories/ScreenGridMap.story.js
```diff
@@ -1,83 +1,99 @@
...
-  const cellSizeOptions = {
-     range: true,
-     min: 1,
-     max: 100,
-     step: 1,
-  };
-  const cellSize = number('Cell Size:', 10, cellSizeOptions);
+          const optionsStyle = {
+            'Hack Oregon Light': 'mapbox://styles/hackoregon/cjiazbo185eib2srytwzleplg',
+            'Hack Oregon Dark': 'mapbox://styles/hackoregon/cjie02elo1vyw2rohd24kbtbd',
```
DingoEatingFuzz: Neat!

mendozaline:
Looks like this solved the GitHub pages issues. Currently, I'm working on doing the same thing for the CivicSandboxMap component.

packages/component-library/stories/ScreenGridMap.story.js
mendozaline: That's what I thought too, but when I didn't include that I kept getting an error that `this` was undefined.

PR #266, opened 2018-06-15
Pie Chart - synchronize props, styling changes, and half doughnut

jaronheard:
The following changes for Pie Chart:
**BREAKING CHANGE**: Changed prop names to match the pattern for our other components
Implemented ChartContainer
Added a half-doughnut option
Default to our Civic categorical colors
Larger text

image

@mendozaline you should be able to use this for the Sandbox dashboard, but let me know if any issues.


jaronheard:
@mendozaline Thanks for the review!

The commits I pushed add a width and height props, that should help you resize I think, but let me know if there's anything else holding it back.

mendozaline:
I think the chart still looks too small, but let's merge it in and go from there.

Issue #192, opened 2018-06-01
Neighborhoods Through The Ages


ylee22:
# Story Card Request

**Project: Neighborhoods**
**Card Title: Neighborhood occupancy for each age group**

## Description / Wireframe

Chart Title: Who lives in [insert neighborhood name]?
Chart Subtitle: Age Distribution of Residents from 2006 to 2016

For each neighborhood with data, there will be a line plot of percent occupancy for each age group. The x-axis will contain years from 2006 to 2016 with increments of 1 year, and the y-axis will be percent occupancy for each age group. There will be 4 different colored lines, each representing the 4 different age groups.

## Data Visualization / Design Concerns
Users will be able to select a neighborhood and visualize a line plot of age group occupancy over year (2006 - 2016 with 1 year increment).

## Context
This will show whether some age groups are getting pushed out of a certain neighborhood.

## Data Details (APIs)
API details to be updated.

Concerns: There might be missing data (not all age groups might have data for a given year and a given neighborhood). If data is missing for a particular year, it will be represented by missing data column. The values for the missing data column will be binary value, such that 0 = no data for that year, 1 = data for that year. There will be missing data column for each age group (4). However for visual continuity, the missing data will be interpolated by averaging between the values on either side (given that the gap is only 1 year).

Dataset: 
Fields: year (x-axis), age group 1 (y-axis), age group 2 (y-axis), age group 3 (y-axis), age group 4 (y-axis)

### How did we calculate this
>Description or link to GitHub description of data methodology

### See the source
>Link to API / data documentation on GitHub
>Link to original source data

jaronheard:
@ylee22 when you produce your table with these values, could you also calculate the percentage change from 2006-2016 for each neighborhood? We could use that to make a layer for the sandbox showing what neighborhoods have people have moved into and out of.

hassanshamim:
http://service.civicpdx.org/neighborhood-development/api/neighborhood_ages is the URL for this data.

hassanshamim:
Convert to total count of voters instead of overall percentage.

lower priority:
- all of portland's numbers

hassanshamim:
add endpoints for:
list of available neighborhoods,
endpoint for filtering by neighborhood

hassanshamim:
@jaronheard since you can get all the data in a single API call (response is only about 300K) do you need separate endpoints for the unique neighborhood names and filtering by endpoints?  If not this should be `API ready`

hassanshamim:
[...]

PR #275, opened 2018-06-17
Card/neighborhoods through the ages

jaronheard:
This is Card #192.

image

Notes: 
Dropdown component in the component library needs to be styled at the component level - I pulled in the CSS used in last year's housing project to test it, but decided to leave it ugly for now and put the CSS in the Dropdown in separate PR.

This could actually be a stacked area, but I tried using the StackedAreaChart component, which should take the same props as the LineChart, and I was getting errors.

Potential improvements:
Setting a default somewhere other than the API reducer - I feel like there's a better spot for this.
Could use some API mocking tests... I tried to write these the other day and after struggling for a while,  I gave up to move on to other things.


jaronheard:
@DingoEatingFuzz I've updated based on your suggestions today, if you want to take a quick look.

[ ] Changed the default to use your .then suggestion
[ ] I wrote a ungroupBy function and put it in dataHelpers.

packages/component-library/src/utils/dataHelpers.js
```diff
@@ -0,0 +1,38 @@
+export const formatData = data => [].concat(
+  data.map(obj => ({ type: '18-25', year: obj.year, pct: obj.pct_18_25 })),
+  data.map(obj => ({ type: '26-32', year: obj.year, pct: obj.pct_26_32 })),
+  data.map(obj => ({ type: '33-39', year: obj.year, pct: obj.pct_33_39 })),
+  data.map(obj => ({ type: '40-49', year: obj.year, pct: obj.pct_40_49 })),
+  data.map(obj => ({ type: '50+', year: obj.year, pct: obj.pct_50_plus })),
+);
```
DingoEatingFuzz: Looks like this is unused now. Great work on this helper, looks really handy.

PR #276, opened 2018-06-17
Landing Page City Lookup

nrbernard:
This pull request replaces the zip code lookup with a static list of zip codes, hooks up the city and state images with search results, and cleans up some linting errors.


jaronheard:
My understanding is that this is incorporated into #279 - if correct, someone close this PR.

Issue #87, opened 2018-05-05
Neighborhoods: Map of Campsite Sweeps (Point Data over Time)


mckinneybock:
- [x] Animate map over time
- [x] Needs API that can return sweeps by time

# Story Card Request

**Project: Neighborhoods**
**Card Title: Map of Campsite Sweeps**

## Description / Wireframe

See Issue #102 - these two cards have been combined into one Sandbox card for the neighborhoods story collection (as of 6/14/18).

Interactive Map of campsite sweeps (link to Issue #81) as part of a greater interactive map representing data about neighborhoods. Represent each sweep as a point on the map; have a dial to represent the data by week (this is the most granular option with this dataset), by month, by 6 months, or by 12 months. A scrolling window would be nice (i.e. any 12-month window would be available). 

In particular, an independent map or visualization with this data would be potentially compelling as a 3D graphic map with quantity of campsite sweeps by month changing over time, perhaps by census block. I'm thinking the style of a relief map, with counts by census block or neighborhood raising or lowering based on the time window & quantity.  (This may be another independent card; I'd be happy to split this off into another issue if so).

A fixed movement through this visualization (either by week or by month) to show the movement of houseless camps and un-housed people over time would visually illustrate how often/frequency of how campsites are displaced. For example, several campsites are swept each week in the same location, and a map visualization that captures this would show frequent movement of sites (and people) over time. 

I don't have a map visualization of the point data right now, but can provide a screenshot if needed. Aggregation over neighborhood is available in Issue #85/Issue #86, and census block can be done using the same methods over the census block polygon in the neighborhoods database. There are lat/long values for the campsite sweeps (~1800) over the time period available (10/2016-present).

## Data Visualization / Design Concerns
- Showing the counts of campsites per week visually, to represent how many sweeps per week/month are occurring over time, via a 3D curved topography (raised where there is a campsite or aggregate of campsites over census block). 
- Visualizing locations/hot spots where lots of campsites are located and frequently swept would show up as an undulation (i.e. from one week to the next, the count would rise, then fall, then rise again). This would require more than just mapping counts, but some sort of delta between weeks to illustrate a quantity of campsites/presence of a campsite, followed by a 'sweep' of that site and then another rise to show that the campsite returned. I'd be happy to chat more in comments.
[...]

PR #285, opened 2018-06-18
WIP - Card/mapping urban campsite displacement

jaronheard:
A WIP PR for card #87. Also includes a small fix for Collapsable.

It sets up the infrastructure to build an animated map of campsite sweeps. I'm hoping that someone can take this infrastructure and run with the animation.

Todo:
- [ ] Add animation (currently there is some non-functional timer redux infrastructure)
- [ ] Add neighborhood boundary layer to map

Current status:
image


geografa:
@jaronheard here's a [rough example](http://bl.ocks.org/geografa/92daf413bf25c1f97cd86a842e34c466) showing the sites by `reportdate`. The example uses an expression that can style a circle with a radius value based on the report date. If you wanted to animate this, you could set up an array of values for the time periods you'd want and then use a `setFilter` on the layer's visibility to turn the layer on and off. Expressions can be used to define the values for paint properties (color, size), layout properties (icons), or filters. Because the reportdate is a string, you may need to get every value and use it in the expression or parse it and do matches on just the month.

PR #292, opened 2018-06-19
Adds sandbox component

nicrocs:
This creates a sandbox component and story that gets the sandbox data and passes it along to the sandbox map. This is still a WIP and we are still getting data defaults set up. But the dropdowns for packages, foundations, and slides are set up pretty much as they will be.


packages/component-library/src/Sandbox/Sandbox.js
```diff
@@ -0,0 +1,110 @@
...
+          slideData={slideData}
+          fetchSlideByDate={fetchSlideDataByDate}
+        />
+
+      </div>
+      <div>
+        <BaseMap
+          mapboxToken={mapboxToken}
+          mapboxStyle={mapboxStyle}
+          initialZoom={10.5}
```
nicrocs: pass in as prop?

PR #295, opened 2018-06-19
adds package selector to civic sandbox

nicrocs:
WIP on setting up the package selector on the civic sandbox. 

- adds redux setup and fetches and sets sandbox state
- displays package selector after sandbox is populated

### DIFF
--- a/packages/civic-sandbox/src/components/App/index.js
+++ b/packages/civic-sandbox/src/components/App/index.js
@@ -1,19 +1,19 @@
 import React from 'react';
 import { css } from 'emotion';
-
-const placeholderClass = css`
-  margin: 30px auto;
-  padding: 30px;
-  border: 2px solid #EEE;
-  text-align: center;
-  max-width: 800px;
-`;
+import Packages from '../Packages';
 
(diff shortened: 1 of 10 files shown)


DingoEatingFuzz:
By WIP do you mean I should hold off reviewing because this shouldn't go into master yet, or do you mean it's not done, but it's at least in a stable spot and can be merged?

nicrocs:
I just mean it's not done. I think it's fine to merge as long as we understand that it's still in progress and doesn't work at this point.

packages/civic-sandbox/src/components/App/index.js
```diff
@@ -1,19 +1,19 @@
 import React from 'react';
 import { css } from 'emotion';
+import Packages from '../Packages';
```
jaronheard: what this does this do? is this necessary? @DingoEatingFuzz

nicrocs: I should probably be clearer with my naming of that component. I named the component for the sandbox package selector `Packages`. Should probably name it `PackageSelector` or `SandboxPackageSelector`

PR #316, opened 2018-06-21
Magnitude of Urban Campsite Sweeps

accua:
Hello everyone, due to a life event, I won't be able to wrap this card up. Right now I'm not getting an x-axis. I think this is because of the `report_time` on the objects returned from the endpoint being a string, rather than date.


jaronheard:
#347 addressed.

PR #317, opened 2018-06-21
Neighborhood Development Card: Class Size and Quality

### LINKED ISSUE #227: Class Size and Quality
- [x] Needs API endpoints

Description / Wireframe

Scatterplot of average teacher experience (years) to average number of student a teacher has in a classroom. There is a dropdown a to select years.

The scatterplot will show data for the selected year:
Note, there will be 4 categories, see details in comments below.
image

Data Visualization / Design Concerns

Having the various schools types in different colors (elementary,PK-5, PK-8/PK-12,6-8,9-12) allows the user to see how staffing is affected by the age of the schoolchildren.

Average class size was used as a metric to determine which schools would receive or lose FTE in the 2018-19 fiscal year. Understanding fluctuations in student teacher ratios across schools and time as well as in relation to teacher experience, provides the end user with context to understand a staffing model

 Data Details (APIs)

pps_classsize.csv (average class size)
school_fte_2018.csv (school_fte_2018_19)

About this data
This data is pulled from published PPS reports.

How did we calculate this?

nrbernard:
Closes #227

Initially, I made the data fetch call with the `year` parameter to only return relevant results. I noticed, however, that other cards fetch much more than the 938 total results for this endpoint and the graphs look better transitioning between year states with all the data already leaded, so now it just fetches it all. If we'd prefer to go the single year fetch route, I can easily change it back.

Also, I'm happy to add tests but just wanted to get this in front of some eyes as soon as possible.


packages/2018-neighborhood-development/src/state/api-adapter.js
```diff
@@ -5,8 +5,8 @@ import axios from 'axios';
 
 const apiAdapter = (url, { start, success, error }) => () => (dispatch) => {
   dispatch(start());
-  return axios.get(url).then((res) => {
-    dispatch(success(res));
+  return axios.get(url).then(({ data }) => {
+    dispatch(success(data.results));
```
nrbernard: Do we mind making this change? I noticed that we were storing the entire response in the Redux store, which seemed unnecessary.

nrbernard:
I'll fix those broken tests a bit later tonight...

packages/2018-neighborhood-development/src/state/api-adapter.js
jaronheard: This is definitely better, but this might be breaking other things to do this since other neighborhood cards use this and are expecting to be getting the data out of the result. I think the better pattern is one that @DingoEatingFuzz set up in the 2018-housing-card with the fetch-adapter. If there was more time I would be cleaning up the code to use that pattern - this API adapter was really me fumbling to get something working, and now that I've worked with it more, I know there are better ways.

packages/2018-neighborhood-development/src/components/App/index.js
```diff
@@ -17,15 +17,15 @@ import DiveDeeperIntoNeighborhoodData from '../DiveDeeperIntoNeighborhoodData';
 
 const App = () => (
   <PageLayout>
+    <ClassSizeAndQuality />
```
jaronheard: I know why you did this, but now you have to put it back 😉

nrbernard: Bahaha I got tired of scrolling!

PR #323, opened 2018-06-21
Use react-dimensions to get BaseMap to be responsive 🗺️

### LINKED ISSUE #290: Figure out Map Width/Height Sizing in a Card
Currently our Maps are coded at a certain percentage of the viewport. This is unideal because when Maps are put inside of a Card, they don't respond to the width of the card. Ideally, there would be a wrapper div around the `BaseMap` component so that then the map will fill up to the wrapper div which scales to the card.

danieltomasku:
This PR adds the `react-dimensions` package which gives the dimensions of a parent container. This is needed because `react-map-gl` width and height props require a number, not percentage or units. Now `BaseMap` can be responsive in a card!

Closes #290


DingoEatingFuzz:
Looks like the BaseMap tests have to be updated now that the map is wrapped in the Dimensions thing. I'm happy to drive this home if you're busy today.

danieltomasku:
@DingoEatingFuzz yeah i'm pretty busy today, feel free to drive it home if you have time! i think this is the related issue: https://github.com/digidem/react-dimensions/issues/35 but feel free to solve the best way you see fit.

DingoEatingFuzz:
@danieltomasku Turns out Dimensions won't render its children if it doesn't have a width and a height...which it doesn't in tests. So I just skipped them 🙃

PR #326, opened 2018-06-21
Feature/adds civic sandbox

nicrocs:
Not pretty, and there's still a bit of work to be done on styles, but this should get us most of the way there.


packages/civic-sandbox/src/state/fetch-adapter.js
```diff
@@ -1,19 +1,46 @@
...
+    return res;
+  }).catch((err) => {
+    dispatch(failure(err));
+  });
+};
+
+export const fetchAllSlidesAdapter = (slides, { start, success, failure }) => () => (dispatch) => {
+  dispatch(start());
+  const fullUrls = slides.map(slide => axios.get(slide.endpoint));
+  return axios.all(fullUrls).then(axios.spread((...res) => {
```
DingoEatingFuzz: Neat! I need to look more into the full axios API. It's got all sorts of stuff I've never seen.

packages/civic-sandbox/src/state/layerStyles.js
```diff
@@ -0,0 +1,1319 @@
...
+      getColor: f => [255, 178, 38, 255],
+      getRadius: f => 100,
+      radiusScale: 1,
+      radiusMinPixels: 1,
+      autoHighlight: true,
+      highlightColor: [100, 100, 100, 100],
+      parameters: { depthTest: false },
+    },
+  },
+});
```
DingoEatingFuzz: This looks like it was a lot of work 😨

PR #332, opened 2018-06-21
updates selector to get packages directly

nicrocs:
Not sure why the old way wasn't working, as it is fine locally, but trying use a selector for packages will hopefully fix it.

-- update. the state wasn't being added to the 2018 package. and after it was added, selectors needed to be updated to look at the correct level

### DIFF
--- a/packages/2018/src/index.js
+++ b/packages/2018/src/index.js
@@ -75,6 +75,7 @@ const configureStore = (initialState, history) => {
       neighborhood: NeighborhoodReducers(),
       transportation: TransportationReducers(),
       farmersMarkets: FarmersMarketsReducers(),
+      sandbox: SandboxReducers(),
     }),
     initialState,
     composeEnhancers(applyMiddleware(...middlewares))
--- a/packages/civic-sandbox/src/state/sandbox/selectors.js
+++ b/packages/civic-sandbox/src/state/sandbox/selectors.js
@@ -12,12 +12,12 @@ export const getState = createSelector(
 
 export const getSandbox = createSelector(
   rootState,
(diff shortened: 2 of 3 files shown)


packages/2018/src/index.js
```diff
@@ -75,6 +75,7 @@ const configureStore = (initialState, history) => {
       neighborhood: NeighborhoodReducers(),
       transportation: TransportationReducers(),
       farmersMarkets: FarmersMarketsReducers(),
+      sandbox: SandboxReducers(),
```
DingoEatingFuzz: Sorry I missed this spot 🙈

packages/civic-sandbox/src/state/sandbox/selectors.js
```diff
@@ -12,53 +12,54 @@ export const getState = createSelector(
...
 
-export const isSandboxLoading = getProperty('sandboxPending');
-export const isFoundationLoading = getProperty('foundationPending');
-export const areSlidesLoading = getProperty('slidesPending');
-export const getSandboxError = getProperty('sandboxError');
-export const getSandboxData = getProperty('sandbox');
-export const getFoundations = getProperty('foundations');
-export const getFoundationError = getProperty('foundationError');
-export const getSelectedPackage = getProperty('selectedPackage');
+export const isSandboxLoading = getSandboxProperty('sandboxPending');
```
DingoEatingFuzz: Not that it matters, but `getSandboxProperty` is now doing what `getProperty` was intended to do. /shrug

nicrocs: yeah, could definitely use some clean up.

PR #339, opened 2018-06-21
Sandbox dashboard

DingoEatingFuzz:
This wires up the sandbox dashboard component to the sandbox.


packages/component-library/src/Sandbox/SandboxDrawer.js
```diff
@@ -52,11 +53,11 @@ const SandboxDrawer = ({
         className={css(`
         position: absolute;
         display: block;
-        height: 100%;
-        width: 20%;
+        width: 250px;
         z-index: 100;
         right: 0;
```
nicrocs: not that it's absolutely necessary, but probably could remove this

Issue #235, opened 2018-06-10
Neighborhoods - Understanding Proposed Staffing Changes in Portland Public Schools 


NatalieRaketic:
# Story Card Request

Project: Neighborhood Development
Title: Understanding Proposed Staffing Changes in Portland Public Schools 

## Description / Wireframe

If table (tables suck):

Headers: School Type, School Name, Proposed Staffing Change, Predicted Staffing Change, Predicted Change in Enrollment, Predicted Change in Free and Reduced Lunch, Student Teacher Ratio, Proposed Student Teacher Ratio(maybe?).

### How did we calculate this
>Description or link to GitHub description of data methodology

### See the source
>Link to API / data documentation on GitHub
>Link to original source data

hassanshamim:
is this still a thing?

jethard:
It’s a table, it should be generated in the script

hassanshamim:
okay so what's the data source?  name of the csv?  I've gotten a lot of csvs I don't remember which this one was.

Is proposed student teacher ratio included in the table?

jethard:
Sorry, I thought i set it up as an output of the notebook, but I guess
that’s not actually helpful to you. I’ll upload the csv to the data folder
and link it in this thread

jethard:
The file is in this folder, under the name dataresults.csv

https://drive.google.com/open?id=1WWhDVlVU6CdfY-xSNwu_a2v_EqMnCJB0

Sorry to make that more complicated then it needed to me.

-Jeanette

hassanshamim:
@jaronheard this is done as well.

Original csv headers:
```csv
School,Students per FTE 2018 - 19,Predictions,Variation,Historically Underserved % ,% Changes in Enrollment,2018-19 % Free Meals by Direct Cert
```

example response:
```json
        {
            "school": "Creative Science",
            "students": 17.5,
            "predictions": 16.916784,
            "variation": 0.583215998,
            "historically_underserved_pct": 0.14,
            "changes_in_enrollment_pct": 0.0,
            "free_meals_direct_cert_pct": 0.14
        }
```


URL: http://service.civicpdx.org/neighborhood-development/api/school_staffing_changes?limit=80

jaronheard:
#354 addressed

Dianna:
@jaronheard , moved to https://github.com/orgs/hackoregon/projects/9 ?

Dianna:
[...]

PR #354, opened 2018-06-29
Card/understanding staff cuts

jaronheard:
This addresses #235 

screen shot 2018-06-28 at 6 20 36 pm

Think this is good enough to merge, but a few questions/comments:
@jethard how do these header labels look to you?
@dsdemaria This uses the DataTable component! Which is sweet! However, there appears to be a bug either in my code or in the DataTable component that is leading to a few values not being displayed. Any chance that you could pull down this branch and check it out, and make an issue if it's related to the component?

### DIFF
--- a/packages/2018-neighborhood-development/src/components/UnderstandingStaffCuts/index.js
+++ b/packages/2018-neighborhood-development/src/components/UnderstandingStaffCuts/index.js
@@ -1,25 +1,148 @@
 import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
 
-import { CivicStoryCard, Placeholder } from '@hackoregon/component-library';
+import { CivicStoryCard, DataTable, Collapsable } from '@hackoregon/component-library';
+import { monthYear } from '@hackoregon/component-library/src/utils/formatters';
+
+import { fetchUnderstandingStaffCuts } from '../../state/understanding-staff-cuts/actions';
+
+import {
+  isUnderstandingStaffCutsPending,
+  catchUnderstandingStaffCutsErrors,
+  getUnderstandingStaffCutsData,
+} from '../../state/understanding-staff-cuts/selectors';
 
 export class UnderstandingStaffCuts extends React.Component {
   componentDidMount() {
-    // initialize data here
+    this.props.init();
   }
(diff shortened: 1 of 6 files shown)


jethard:
Headers look good

packages/2018-neighborhood-development/src/components/UnderstandingStaffCuts/index.js
```diff
@@ -1,25 +1,147 @@
...
+              <p>See the results of our analysis in the sortable table below.</p>
+            </div>
+          </Collapsable.Section>
+          <Collapsable.Section hidden>
+            <div>
+              { understandingStaffCuts &&
+                <DataTable
+                  data={tableData}
+                />
+              }
```
DingoEatingFuzz: Seems weird for the data component to be hidden by default.

packages/2018-neighborhood-development/src/components/UnderstandingStaffCuts/index.js
```diff
@@ -1,25 +1,147 @@
...
+            <p>Due to the quality of data and complexity of school populations, Hack Oregon elected to conduct a simple linear regression analysis. Using data from the original PDF and PPS data on underserved populations our analysis concluded on the following model:</p>
+            <p>Students per FTE 2018 - 19 = 18.82 - 2.69*(Historically Underserved %) + 1.58*(% Changes in Enrollment)  - 10.91*(2018-19 % Free Meals by Direct Cert)</p>
```
DingoEatingFuzz: We should work on some latex-esque calculation styles.

packages/2018-neighborhood-development/src/components/UnderstandingStaffCuts/index.js
jaronheard: Yeah, it's definitely a bit off. It was a workaround to avoid having to implementing pagination or some other method of limiting the length the DataTable component, because as is it's something like three screens long, which seemed a little much for a single card.

Perhaps that's something that could be implemented in collapsable - something like a max height for a section that would cut off mid-component. That would require some creative reworking of collapsable, and I'm not sure if that's the right approach.

Issue #102, opened 2018-05-06
Explore Urban Campsite Sweeps by Neighborhood / Human Impact


mckinneybock:
# Story Card Request

**Project: Neighborhoods**
**Card Title: Explore Urban Campsite Sweeps by Neighborhood**

## Description / Wireframe

Potentially contains an animation by week or month aggregate, that flashes through all available sweeps data from 10/2016-present. (April 2018, as May 2018 is incomplete).

Also ideally an interactive map of campsite sweeps (by point), with neighborhood boundaries, to illustrate how many sweeps there have been. This will help show where sweeps are happening around the city; it will be combined with descriptive statistics (means) on estimated numbers of people per campsite to estimate the number of people affected per week in the Expanded Content.

Related to Issue #100 

## Data Visualization / Design Concerns
Animation over time is very important
Second is an interactive component, so users can explore their neighborhoods & areas of Portland over time.

## Context
This is a pretty critical comparison to make to better understand how frequent the campsite sweeps are occurring, and what response is to the local un-housed population.

## Data Details (APIs)

datasets:

- campsite sweeps

Data will be located in the Neighborhoods API.

### How did we calculate this
>Description or link to GitHub description of data methodology

### See the source
>Link to API / data documentation on GitHub
>Link to original source data

christinatommeraasen:
Neighborhood map on one side and a line graph on the other.

jaronheard:
This is in the Civic Sandbox

PR #356, opened 2018-06-30
Magnitude of urban campsite animation

jaronheard:
Adds animation to #102

### DIFF
--- a/packages/2018-neighborhood-development/src/components/ExploreUrbanCampsiteSweeps/index.js
+++ b/packages/2018-neighborhood-development/src/components/ExploreUrbanCampsiteSweeps/index.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { css } from 'emotion';
 
-import { CivicStoryCard, Collapsable, BaseMap, ScatterPlotMap, ChartContainer } from '@hackoregon/component-library';
+import { CivicStoryCard, Collapsable, BaseMap, ScatterPlotMap, Button, ChartContainer } from '@hackoregon/component-library';
 import { contextualDesc, belowFoldOne, belowFoldTwo } from './text';
 
 import { fetchCampsiteSweeps, incrementTimer } from '../../state/explore-urban-campsite-sweeps/actions';
@@ -11,6 +11,7 @@ import {
   isCampsiteSweepsPending,
   catchCampsiteSweepsErrors,
   getCampsiteSweepsData,
+  getCampsiteSweepsTimer,
 } from '../../state/explore-urban-campsite-sweeps/selectors';
 
 const LAT = 45.5231;
@@ -27,9 +28,9 @@ export class ExploreUrbanCampsiteSweeps extends React.Component {
       isLoading,
       error,
       data,
+      timer,
+      play,
     } = this.props;
-
-    console.log('DATA???', data);
 
     const scatterplot = !data
     ? null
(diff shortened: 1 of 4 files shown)


packages/2018-neighborhood-development/src/components/ExploreUrbanCampsiteSweeps/index.js
```diff
@@ -77,24 +79,25 @@ export class ExploreUrbanCampsiteSweeps extends React.Component {
...
+    data: getCampsiteSweepsDataByTime(state),
+    timer: getCampsiteSweepsDateRange(state),
   }),
   dispatch => ({
     init() {
       dispatch(fetchCampsiteSweeps());
-    },
-    incrementTimer() {
-      dispatch(incrementTimer(5));
+      setInterval(() => dispatch(incrementTimer(1)), 800);
```
DingoEatingFuzz: Calling `setInterval` like this may cause a memory leak. I suspect you will have to clear the interval when the component is unmounted. Something like...

```js
class Card extends React.Component {
  componentDidMount() {
    const timer = setInterval(() => { this.props.tick(); }, 800);
    this.setState({ timer });
  },

  componentWillUnmount() {
    clearInterval(this.state.timer);
  },
}

export default connect({
  state => ...,
  dispatch => ({
    tick() {
      dispatch(incrementTimer(1));
    }
  })
});
```

This means the component is stateful but at least only the very local timer state is stored on the component instead of in the Redux store. Other possibilities: https://medium.com/@machadogj/timers-in-react-with-redux-apps-9a5a722162e8

packages/2018-neighborhood-development/src/state/explore-urban-campsite-sweeps/selectors.js
```diff
@@ -20,3 +22,51 @@ export const catchCampsiteSweepsErrors = createSelector(
...
+  const startMonth = min(timepoints).getMonth();
+  const startYear = min(timepoints).getFullYear();
+  const endMonth = max(timepoints).getMonth();
+  const endYear = max(timepoints).getFullYear();
+  const startDate = new Date(startYear, startMonth);
+  const endDate = new Date(endYear, endMonth, 0);
+  const months = (endDate.getMonth() - startDate.getMonth()) + 1 + ((endDate.getFullYear() - startDate.getFullYear()) * 12);
+  const sections = [...Array(months + 1).keys()];
+  return sections.map(item => new Date(startDate.getFullYear(), startDate.getMonth() + item));
+};
```
DingoEatingFuzz: Oof. Date stuff in vanilla js is so hairy. It might be time to pull in a date lib. I'm thinking maybe [date-fns](https://date-fns.org/) instead of the goto-pick of Moment.

packages/2018-neighborhood-development/src/components/ExploreUrbanCampsiteSweeps/index.js
jaronheard: Did essentially this, you're only storing an ID of the interval you set in state, so the state is very minimal. Thanks for the help.

packages/2018-neighborhood-development/src/state/explore-urban-campsite-sweeps/selectors.js
jaronheard: Was thinking of creating a new issue for bringing in date-fns rather than addressing this right now - I'd like to clean up some of the other date code and add it to the documentation while I'm at it.

jaronheard:
@DingoEatingFuzz I think this is ready for re-review and merge unless you'd like the vanilla JS dates to use date-fns first

Issue #212, opened 2018-06-07
The Challenge of Preserving Affordability in a Complex Housing Market


danmelton:
# Story Card Request

**Project: Housing**
**Card Title: The Challenge of Preserving Affordability in a Complex Housing Market**

## Description / Wireframe

We'd like to embed the Metro housing pdf as an image on the page, and when clicked, have it get bigger (perhaps a modal?)

https://www.oregonmetro.gov/sites/default/files/2018/04/19/Equitable-Housing-Initiative-Factsheet-Affordability-201804.pdf

it could also possibly be an iframe?

jaronheard:
#358 addressed

PR #358, opened 2018-07-04
Card / The Challenge Of Preserving Affordability In A Complex Housing Market

jaronheard:
This addresses #212.

I added a PDF component to the component library for this card. I'm sure there is opportunity to improve that component, in terms of appropriately sizing the PDF container and possibly the implementation of the PDF container, but this seems to work for now.

screen shot 2018-07-03 at 5 06 52 pm

### DIFF
--- a/packages/component-library/src/PDF/PDF.js
+++ b/packages/component-library/src/PDF/PDF.js
@@ -3,7 +3,7 @@ import { css } from 'emotion';
 
 const pdfClass = css`
   width: 100%;
-  height:calc(100vh - 70px);
+  height: 700px;
 `;
 
 const PDF = ({ url }) => (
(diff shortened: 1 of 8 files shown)


packages/component-library/src/PDF/PDF.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { css } from 'emotion';
+
+const pdfClass = css`
+  width: 100%;
+  height:calc(100vh - 70px);
```
DingoEatingFuzz: Using `vh` here seems weird. What's the goal with this?

jaronheard: Was trying to make it so that the PDF embed fills most of, but not all of the screen on desktop and mobile. There’s probably a better way to do this. Will revisit.

PR #366, opened 2018-07-12
Sandbox Layer Improvements

mendozaline:
As I was going through the different packages in the sandbox, I noticed some issues that I fixed here.
- [x] Changed the color scheme for the transportation package from red/green to purple/green since some people have [problems distinguishing between red and green](https://nei.nih.gov/health/color_blindness/facts_about). 
- [x] Fixed bike count and bike estimate slide layers in the bike package that weren't displaying because of null geometries
- [x] Scaled bike count and bike estimates bubbles by area instead of radius
- [x] Decreased the line width of the parks and community garden polygons 

There are some other smaller changes like deleting console.logs and other code clean up.


jaronheard:
Looks good! 🚀

My only comment is to consider changing the slide color for crashes in the transportation package from purple to red - with the change to purple-green they blend in a little too much. But not going to hold off on these other improvements for that.

mendozaline:
Changed the crash circle color to red. That was a good suggestion. It's definitely more intuitive for crashes to be red anyway.

PR #369, opened 2018-07-17
Card/votersonthemove

jaronheard:
Another story card!

This one uses a "small multiples" layout.

image


DingoEatingFuzz:
That table looks so sophisticated 📚

packages/2018-neighborhood-development/src/state/voters-on-the-move/selectors.js
```diff
@@ -0,0 +1,37 @@
...
+);
+
+export const isAwayVotersOnTheMovePending = createSelector(
+  getVotersOnTheMoveRequest,
+  ({ awayPending }) => !!awayPending,
+);
+
+export const catchAwayVotersOnTheMoveErrors = createSelector(
+  getVotersOnTheMoveRequest,
+  ({ awayError }) => awayError || awayError,
```
DingoEatingFuzz: 🤔

packages/2018-neighborhood-development/src/state/voters-on-the-move/selectors.js
```diff
@@ -0,0 +1,37 @@
...
+);
+
+export const isVotersOnTheMovePending = createSelector(
+  getVotersOnTheMoveRequest,
+  ({ pending }) => !!pending,
+);
+
+export const catchVotersOnTheMoveErrors = createSelector(
+  getVotersOnTheMoveRequest,
+  ({ error }) => error || error,
```
DingoEatingFuzz: 🤔

jaronheard:
@DingoEatingFuzz I think that the issue with the scatterplots is one of inadequate labelling and description, and it made me think about the information hierarchy on this card. I ended up adding chart titles and subtitles to both of the visualizations in this card, for which I separated a ChartTitle component.

untitled

Does that help address the scatterplot clarity? As an aside, I think that generally these are confusing charts and if I was to redesign this card I would do it differently.

packages/2018-neighborhood-development/src/state/voters-on-the-move/selectors.js
```diff
@@ -0,0 +1,37 @@
+import { createSelector } from 'reselect';
+import { rootState } from '../selectors';
+
+export const getVotersOnTheMoveRequest = createSelector(
+  rootState,
+  ({ votersOnTheMove }) => votersOnTheMove,
+);
+
+export const getVotersOnTheMoveData = createSelector(
```
nrbernard: I'm sure it's a pattern we're using throughout the app, but it seems like overkill to use selectors here. Since we're not computing any derived data from state, I don't think there's any benefit and it makes it a little more confusing than just doing something like this in our `mapStateToProps` functions:

```
export default connect(
  state => ({
    votersOnTheMove: state.neighborhood.votersOnTheMove,
    ...
  }),
);
```

I know we're composing them with `getVotersOnTheMoveRequest ` to dry up the code but that's not computing anything either. I'm sure I'm missing something, though, so just curious!

packages/2018-neighborhood-development/src/components/VotersOnTheMove/index.js
```diff
@@ -1,25 +1,239 @@
...
+      error,
+      votersOnTheMove,
+      isAwayLoading,
+      awayError,
+      awayVotersOnTheMove,
+    } = this.props;
+
+    const data = { center: votersOnTheMove, away: awayVotersOnTheMove };
+
+    function legendComponent(age) {
```
nrbernard: This is just a style thing but I feel like we should move this out of the `render` method. We could simplify the function signature to something like `const legendComponent = (age) => (lengendData) => ...` just to make it clear that we're currying here.

jaronheard: Actually had the same thought when I was looking at this code earlier... So now that there's two voices saying it's weird I think I'm obligated to fix it 😝

jaronheard: Actually had the same thought when I was looking at this code earlier... So now that there's two voices saying it's weird I think I'm obligated to fix it

packages/2018-neighborhood-development/src/state/voters-on-the-move/selectors.js
jaronheard: I'll bet @DingoEatingFuzz has some thoughts on this pattern, which I'd be interested to hear as well!

DingoEatingFuzz: I do have thoughts! You are correct, @nrbernard,  that `getVotersOnTheMoveRequest` itself isn't doing much, just peeling a property off of the response, but it is built on top of the `rootState` selector that conditionally uses the true root of state or the scope project root of state (i.e., `root.neighborhoodDevelopment`). 

By skipping selectors and doing the quick property lookup in the component as you suggest, you have to increase the surface area where we have to consider the store merging that happens when running the 2018 package vs. the neighborhood-development package in isolation.

All that said, I'm not opposed to rethinking our approach to this, but I do think it's nice that the store merging thing is abstracted to the point where it can essentially go unnoticed.

More on this in [ARCHITECTURE.md](https://github.com/hackoregon/civic/blob/master/ARCHITECTURE.md#creating-a-unified-redux-store).

DingoEatingFuzz:
@jaronheard I don't think the labels helped me at all. It was a good attempt, but I'm not grokking the data encoding here.

First, what are the units for the axes? Are they miles?

Second (assuming miles, or at least distance), is that number of miles _moved_? In other words, a vector. Or is it the coordinates the person moved to? In other words a point.

All four charts have a cluster in the center which seems to contradict with the titles of the columns. If these are movements _away_ from the city center, why are the points still clustered in the center?

To make the chart work, I am assuming that a point represents a vector. Meaning the points in the center of the away charts mean those people just didn't move very far away. Likewise for the towards charts: they didn't move very far towards city center.

If this is the case, I think the charts are both hard to interpret and not very insightful (the insight being that various people moved various distances). 

> As an aside, I think that generally these are confusing charts and if I was to redesign this card I would do it differently.

👍 I'm not suggesting a redesign either, just going through what is hopefully a constructive review from fresh eyes.

jaronheard:
@DingoEatingFuzz Good points, and constructive.

Units are miles moved from original location to new location - so they represent a vector. Because the dataset is bounded by the city limits, the possibility for long distance moves is limited. In addition, most people who are moving within Portland aren't moving very far. Which is at least some kind of insight.

I made another effort at explaining what we have, which I think is a slight improvement. Calling the charts *Various People Moved Various Distances* would actually be pretty accurate. Maybe I should just be using that.

Couple of potential improvement options...

image
image

PR #374, opened 2018-07-26
Fix Sandbox Map Tooltip

mendozaline:
Currently the tooltip for the sandbox uses the field names from the property object for its text. So some of the descriptions are not very useful.

I finally got around to fixing that so the text is properly formatted and uses the field names provided in `slide_meta.attributes` for each slide.

One issue that I encountered was that some of the `slide_meta.attributes` fields don't match the names of the fields in the actual data. For example, in [slide 016 - points of interest](http://service.civicpdx.org/neighborhood-development/sandbox/slides/retailgrocers/?format=json), the value for `slide_meta.attributes.secondary.field` is `description_txt` but the field name in `slide_data.features.properties` is `description2_txt`. So that mismatch prevents the tooltip from showing up since they don't match. The same problem happens with [slide 015 - change in ridership by route](http://service.civicpdx.org/transportation-systems/sandbox/slides/routechange/?format=json).

Another issue is that some slides have information in `slide_data.features.properties`, but there are no `primary` or `secondary` fields provided in `slide_meta`. [Slide 011 - demolitions](http://service.civicpdx.org/neighborhood-development/sandbox/slides/demolitions/) and [slide 032 - crashes](http://service.civicpdx.org/transportation-systems/sandbox/slides/crashes/) share that issue.

### DIFF
--- a/packages/civic-sandbox/src/state/sandbox/selectors.js
+++ b/packages/civic-sandbox/src/state/sandbox/selectors.js
@@ -180,7 +180,7 @@ export const getSelectedSlideDatum = createSelector(
     const tooltipObj = {};
     tooltipObj['x'] = selectedSlideDatum.x;
     tooltipObj['y'] = selectedSlideDatum.y;
-    tooltipObj['content'] = {};
+    tooltipObj['content'] = [];
 
     const tooltipSlideName = Object.keys(slide[slideIndex]);
 
@@ -191,10 +191,16 @@ export const getSelectedSlideDatum = createSelector(
     const datumProps = selectedSlideDatum.object.properties;
 
     if (tooltipPrimary && tooltipPrimary.field) {
-      tooltipObj.content[tooltipPrimary.name] = datumProps[tooltipPrimary.field];
+      tooltipObj.content.push({
+        name: tooltipPrimary.name,
+        value: datumProps[tooltipPrimary.field],
+      });
     }
     if (tooltipSecondary && tooltipSecondary.field) {
-      tooltipObj.content[tooltipSecondary.name] = datumProps[tooltipSecondary.field];
+      tooltipObj.content.push({
+        name: tooltipSecondary.name,
+        value: datumProps[tooltipSecondary.field],
+      });
     }
 
     return tooltipObj;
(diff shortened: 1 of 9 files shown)


packages/component-library/src/CivicSandboxMap/CivicSandboxTooltip.js
```diff
@@ -17,31 +17,26 @@ const tooltip = css`
...
 
-  const keyValuePairs = Object.entries(tooltipInfo.properties);
-
-  const tooltipContent = keyValuePairs.map((d, i) => {
-    const nameSplit = d[0].split("_");
-    const name = nameSplit.reduce((acc, cur, i) => {
-      const capitalize = cur.charAt(0).toUpperCase() + cur.slice(1);
-      return acc + " " + capitalize;
-    }, "");
+  const keyValuePairs = Object.entries(tooltipData.content);
```
jaronheard: Is there a better name than keyValuePairs for this variable?

packages/component-library/src/CivicSandboxMap/CivicSandboxTooltip.js
```diff
@@ -52,20 +47,10 @@ const MapTooltip = (props) => {
...
-      { tooltipRender }
-    </div>
-  );
 };
 
 MapTooltip.propTypes = {
-  tooltipInfo: PropTypes.object,
-  x: PropTypes.number,
-  y: PropTypes.number,
+  tooltipData: PropTypes.object,
```
jaronheard: Maybe adding an object shape here could assist in reasoning about what keyValuePairs is doing?

packages/civic-sandbox/src/state/sandbox/selectors.js
```diff
@@ -142,3 +142,61 @@ export const getSelectedFoundationDatum = createSelector(
...
+    if (tooltipPrimary && tooltipPrimary.field) {
+      tooltipObj.content[tooltipPrimary.name] = datumProps[tooltipPrimary.field];
+    }
+    if (tooltipSecondary && tooltipSecondary.field) {
+      tooltipObj.content[tooltipSecondary.name] = datumProps[tooltipSecondary.field];
+    }
+
+    return tooltipObj;
+  }
+);
```
jaronheard: Slick selector! 🎯

AraOshin:
@mendozaline @jaronheard 

Yes! It's awesome to see this update. I'll look into those api issues you mentioned this week. 👍

Issue #261, opened 2018-06-14
adjust corner branding and title responsiveness for cleaner look on small screens


sixfoot6:
On smaller screens (say, below ~600px or whatever our medium breakpoint is), I'd recommend that we

1. Reduce H2 font-size to 2em
2. Shrink both heigh and width of civid corner to 75px

This, along with the simplification of the card footer actions in case #259, should leave us in good shape on small screens.

![screenshot 2018-06-14 10 10 48](https://user-images.githubusercontent.com/11667/41427370-8c10cfd8-6fbb-11e8-82d4-449b65760c03.png)

jaronheard:
#375 addressed

Issue #254, opened 2018-06-13
Left align chart + add spacing in header container on mobile


jaronheard:
Chart container should render title and subtitle left aligned on mobile. 

![a6eb3656-148e-4157-87fe-5baf858b4a72](https://user-images.githubusercontent.com/7065695/41379712-937d55f6-6f17-11e8-8a2d-4afb72d8f5a0.png)

karieleana:
Add more space between chart title + text....

jaronheard:
#375 addressed

Issue #206, opened 2018-06-06
Scatter plot legend positioning differs between browsers


flamingveggies:
Take a look at: https://hackoregon.github.io/civic/?knob-Data%20key=sortOrder&knob-Title=Some%20title&knob-Data=%5B%7B%22x%22%3A1%2C%22y%22%3A2%2C%22series%22%3A%22cat%22%7D%2C%7B%22x%22%3A2%2C%22y%22%3A3%2C%22series%22%3A%22cat%22%7D%2C%7B%22x%22%3A3%2C%22y%22%3A5%2C%22series%22%3A%22fish%22%7D%2C%7B%22x%22%3A4%2C%22y%22%3A4%2C%22series%22%3A%22fish%22%7D%2C%7B%22x%22%3A5%2C%22y%22%3A7%2C%22series%22%3A%22cat%22%7D%2C%7B%22x%22%3A1%2C%22y%22%3A3%2C%22series%22%3A%22dog%22%7D%2C%7B%22x%22%3A3%2C%22y%22%3A3%2C%22series%22%3A%22dog%22%7D%5D&knob-Data%20values=population&knob-Subtitle=A%20description%20of%20this%20chart.&knob-dataValue=y&knob-yLabel=Rating&knob-xLabel=Number&knob-Data%20Series=%5B%22cat%22%2C%22dog%22%2C%22fish%22%5D&knob-Size=%7B%22key%22%3A%22y%22%7D&knob-Domain=%7B%22x%22%3A%5B0%2C6%5D%2C%22y%22%3A%5B0%2C8%5D%7D&knob-dataKey=x&knob-Data%20key%20labels=label&selectedKind=Scatterplot&selectedStory=with%20props&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

The legend appears centered above the chart in Chrome, and what appears to be left-justified in Safari. I have not tested any other browsers or non-Mac OSs.

jaronheard:
This also impacts the line chart component, which also uses the same legend component.

jaronheard:
#375 addressed

PR #375, opened 2018-07-28
Updates per design QA and legend positioning fix.

jaronheard:
Some CSS styling improvements per #254 and #261.

In adjusting the legend positioning, I also figured out a fix for #206. I switched an unnecessary`<legend>` to a `<div>` in the SimpleLegend component.

### DIFF
--- a/packages/component-library/src/SimpleLegend/SimpleLegend.js
+++ b/packages/component-library/src/SimpleLegend/SimpleLegend.js
@@ -9,11 +9,15 @@ const SimpleLegend = ({ legendData }) => {
     font-weight: bold;
     text-align: center;
     margin: 10px 0 0 0;
+
+    @media (max-width: 640px) {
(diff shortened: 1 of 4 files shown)


packages/component-library/src/SimpleLegend/SimpleLegend.js
```diff
@@ -9,11 +9,15 @@ const SimpleLegend = ({ legendData }) => {
...
+
+    @media (max-width: 640px) {
+        text-align: left;
+    }
   `;
 
   if (legendData.length) {
     return (
-      <legend className={legendStyle}>
+      <div className={legendStyle}>
```
DingoEatingFuzz: I imagine there's a default property set for the legend element in the safari user agent styles that could be overridden to make this consistent, but I don't mind the switch to a `div` either.

PR #377, opened 2018-08-11
Collection page redesign

jaronheard:
Simplified design for the Portland collections page that no longer uses the data visualization colors to be able to extend include more than 5 collections. 

**Desktop:**
screenshot_2018-08-11 civic 2018 - a hack oregon project 1
**Mobile:**
screenshot_2018-08-11 civic 2018 - a hack oregon project

### DIFF
--- a/packages/2018/src/components/PortlandCollectionPage.js
+++ b/packages/2018/src/components/PortlandCollectionPage.js
@@ -8,24 +8,16 @@ const collectionPageWrapper = css`
   width: 100%;
   overflow-y: hidden;
   h1 {
-    width: 400px;
-    display: inline-block;
-    font-size: 80px;
+    font-size: 50px;
     line-height: 1.2;
-    letter-spacing: -1px;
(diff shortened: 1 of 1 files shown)


DingoEatingFuzz:
Hm. I appreciate the design constraint of needing to support an arbitrarily long list, but this page is awfully stark now.

Maybe if the whole page had a background color that the white cards sat on top of?

I'm also not convinced that that is the best treatment of the sandbox link. I like that it gets more visual significance now, but treating it like a project feels incorrect.

I don't want to fuss the design details too much, but someone ought to review them 😄

jaronheard:
@DingoEatingFuzz I pushed an update to add a link to the API endpoint collection page!

I agree with you a bit about the design being a bit stark, but talking to Cat she said that this was very close to what the design team had envisioned. With her approval, I'd like to get this one merged.

I can see this design changing or being iterated on here soon, but I think in the context of our other pages, it actually fits pretty well.

Issue #269, opened 2018-06-16
Civic Sandbox Dashboard


mendozaline:
The Civic Sandbox map will include a dashboard as seen in issue [#163](https://github.com/hackoregon/civic/issues/163).

The dashboard will display data from the foundation layers using the [half-doughnut chart](https://github.com/hackoregon/civic/issues/245), [comparison bars](https://github.com/hackoregon/civic/issues/245), [gradient scale components](https://github.com/hackoregon/civic/issues/218), and text.
![Dashboard1](https://i.imgur.com/4v00DnN.png)

It will also include a view that provides a description of the map. Users can switch back and forth between the data display and the description.
![Dashboard2](https://i.imgur.com/skd4mSv.jpg)

jaronheard:
#271 addressed

PR #382, opened 2018-08-26
Update Civic Sandbox Dashboard

mendozaline:
This is a follow up on [#269](https://github.com/hackoregon/civic/issues/269). 

The dashboard now includes the comparison bar and legend options. However, most of the comparison bars have placeholder values. So we can now update those foundation layers with real data. I planned on using the Gradient Scale component for the map legend, but I ran into issues with that. Instead I had to create something new. This meant only showing colors visible on the map and not all 9 colors in the scale. And only displaying the minimum and maximum numbers on the ends. I'm still trying to think of how to show the intermediate values. Also for the disaster response foundations, I had to account for the categorical data.

Other changes:
- [x] Increased the height of the map
- [x] Dashboard buttons only render if a description is provided
- [x] Moved the dashboard to the left to avoid overlap with the slide selector
- [x] Disable zoom on scroll

To do:
- [x] Make the dashboard responsive for small screens

### DIFF
--- a/packages/civic-sandbox/src/components/Packages/index.js
+++ b/packages/civic-sandbox/src/components/Packages/index.js
@@ -112,7 +112,20 @@ export class Packages extends React.Component {
             <a onClick={this.closeMap}>&lt; Back to Packages</a>
           </p>
           <SandboxComponent />
-          {selectedFoundationDatum && <div>
+          {selectedFoundationDatum && <div
+            className={css(`
+              position: absolute;
+              top: 21%;
+              left: 4%;
+              width: 96%;
+              height: 0;
+              @media(max-width: 900px) {
+                position: relative;
+                left: 0;
+                height: auto;
+              }
+            `)}
+          >
             <CivicSandboxDashboard data={selectedFoundationDatum}>
               {/*<div className={css(`display: inline-block; width: 90%; margin: 1% 2% 2% 7.5%;`)}>
                 <h2>How has ridership changed throughout Tri-Met's service area over time?</h2>
--- a/packages/civic-sandbox/src/components/Packages/index.js
+++ b/packages/civic-sandbox/src/components/Packages/index.js
@@ -114,6 +114,14 @@ export class Packages extends React.Component {
           <SandboxComponent />
           {selectedFoundationDatum && <div>
             <CivicSandboxDashboard data={selectedFoundationDatum}>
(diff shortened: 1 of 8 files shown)


mendozaline:
To make the dashboard responsive, I simply changed it so the dashboard now appears below the map on small screens. Similar to the approach used on maps in previous collections.

I also updated the civic sandbox dashboard's story in the storybook.

packages/civic-sandbox/src/components/Packages/index.js
```diff
@@ -113,13 +113,29 @@ export class Packages extends React.Component {
...
+                position: relative;
+                left: 0;
+                height: auto;
+              }
+            `)}
           >
-            <CivicSandboxDashboard data={selectedFoundationDatum} />
+            <CivicSandboxDashboard data={selectedFoundationDatum}>
+              {/*<div className={css(`display: inline-block; width: 90%; margin: 1% 2% 2% 7.5%;`)}>
+                <h2>How has ridership changed throughout Tri-Met's service area over time?</h2>
```
jaronheard: Can this comment be removed?

packages/civic-sandbox/src/state/sandbox/selectors.js
```diff
@@ -110,13 +110,96 @@ const makeVisFor = (spec, data) => {
     };
   }
   if (type === 'Text') {
-    console.log('TYPE', spec.name, spec.field, data.properties[spec.field]);
+    // console.log('TYPE', spec.name, spec.field, data.properties[spec.field]);
```
jaronheard: Still needed?

packages/component-library/src/BaseMap/BaseMap.js
```diff
@@ -32,6 +32,7 @@ class BaseMap extends Component {
         maxZoom: 16,
         pitch: props.initialPitch || 0,
         bearing: 0,
+        scrollZoom: false,
```
jaronheard: On Chrome on MacOS using a trackpad, not having scrollZoom feels inconsistent with the behavior of other interactive maps. If this is targeting other use cases (e.g. mobile), maybe we can selectively target those use cases.

mendozaline: I'll figure this out in a future iteration of the Base Map. I changed it back since it's not a necessary change for the dashboard.

Issue #371, opened 2018-07-19
Neighborhoods Cards (all) - source links broken


mckinneybock:
# Story Card Request

**Project: Neighborhoods**
**Card Title: all**

The Source Links for the neighborhoods projects don't link back to the methods reports we provided in the Google doc. Here is the original document with source links provided for each card, linking back to the github repo folder that contains our methods reports/further documentation: 

https://docs.google.com/document/d/1L5FfnL9JPQV2VCSmK2W3ykk_ZWoGTch-_JHV5nJEYVc/edit

mckinneybock:
@ylee22, @arailsya, @jethard

mckinneybock:
Also, a critical point for THE SANDBOX (potentially @AraOshin, @jaronheard)

The Eviction Lab data is in Sandbox, and it needs to be cited somewhere whenever it’s called up. There is a short blurb that needs to be cited, with the following (taken from their webpage):

Any works citing Eviction Lab data should use the following citation:

This research uses data from The Eviction Lab at Princeton University, a project directed by Matthew Desmond and designed by Ashley Gromis, Lavar Edmonds, James Hendrickson, Katie Krywokulski, Lillian Leung, and Adam Porton. The Eviction Lab is funded by the JPB, Gates, and Ford Foundations as well as the Chan Zuckerberg Initiative. More information is found at evictionlab.org.
If you use this data in your research or reporting, please share it with us.

The “share with us” link heads here: https://evictionlab.org/contact/

Thanks, you all!

jaronheard:
#384 and #385 addressed.

PR #385, opened 2018-09-11
Add source link to Sandbox footer

jaronheard:
This will complete #371 by adding a data sources attribution link to the Sandbox footer.

I see this as a stop gap solution to at least cover our bases for the attribution that we need for the Eviction Labs data. Ideally, the data sources would be connected to each layer, but since we don't have that capability right now, I'm linking to our default data sources page.

### DIFF
--- a/packages/component-library/src/Footer/Footer.js
+++ b/packages/component-library/src/Footer/Footer.js
@@ -73,10 +73,7 @@ const scrollToTopClass = css`
   }
 `;
 
-const defaultAttribution = (
-  <div>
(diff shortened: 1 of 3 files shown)


packages/component-library/src/Footer/Footer.js
```diff
@@ -72,9 +73,14 @@ const scrollToTopClass = css`
   }
 `;
 
-const Footer = () => (
+const defaultAttribution = (
+  <div>
```
nrbernard: Not a big deal at all but this wrapping `div` is unnecessary.

Issue #387, opened 2018-09-15
Add Heat Map Component


mendozaline:
We need a heat map component. To be made with mapbox gl: https://www.mapbox.com/mapbox-gl-js/example/heatmap-layer/

Todo:
- [x] Create HeatMap component
- [x] Add HeatMap story

mendozaline:
Addressed by [#390](https://github.com/hackoregon/civic/pull/390).

Issue #166, opened 2018-05-24
Building Boom in Portland


kmorrice:
# Story Card Request

**Project: Housing 2018**
**Card Title: Building Boom in Portland**

## Description / Wireframe
![ezgif com-video-to-gif](https://user-images.githubusercontent.com/3445877/40461592-e634e6fe-5ec0-11e8-8a18-23247d76fb8e.gif)

We would like to show animations of building permits for single-family housing, multi-family housing, and ADUs. There are data going back to ~1994-1996 for each set, and we would like to feature heat maps that portray the density of permits up through 2017. We would also like to show either a bar chart or a time series indicating the number of building permits for these different categories over time. 

## Data Visualization / Design Concerns
> If you have specific design or data visualization concerns, mention them here

## Context
> Why is this important? This information will be used to prioritize and contextualize story cards.

## Data Details (APIs)

https://opendata.arcgis.com/datasets/4d04eba5e9934a26be530076cae65839_89.geojson

We need to filter on several fields, including NEW CLASS, and NEW TYPE.  

### How did we calculate this
>Description or link to GitHub description of data methodology

### See the source
>Link to API / data documentation on GitHub
>Link to original source data
http://gis-pdx.opendata.arcgis.com/datasets/residential-building-permits

danmelton:
It looks we can use the ScreenGridMap Component in the library for this card

kmorrice:
We also want to include either a bar chart or line chart that could be shown next to its respective animation. It shows the number of permits per year over time. 
![adus](https://user-images.githubusercontent.com/3445877/40857829-2db80552-6591-11e8-9727-9e0cf07e14e8.png)
![multi-family](https://user-images.githubusercontent.com/3445877/40857830-2dcf3a4c-6591-11e8-8f9b-f44435375d98.png)
![single-family](https://user-images.githubusercontent.com/3445877/40857831-2de5ae1c-6591-11e8-88ab-a35d9880fafb.png)

adpeters:
API Details:

We are defining three categories of dwellings, each with a specific filter:

Single-Family: http://service.civicpdx.org/housing-affordability/api/permits/?new_type=Single+Family+Dwelling&new_class=New+Construction&is_adu=false

Multi-Family: http://service.civicpdx.org/housing-affordability/api/permits/?new_type=Apartments/Condos,Duplex,Rowhouse,Townhouse&new_class=New+Construction&is_adu=false

[...]

PR #390, opened 2018-09-16
Issue #387 Heat Map

mendozaline:
This is the heat map component needed for issue #166. I added the component and story. I included some different color and data options to test it out.


mendozaline:
I added some default props and I agree that going forward we should follow the BaseMap and layers pattern. Hopefully deck.gl will add a true heat map layer soon and we can use that instead.

Issue #106, opened 2018-05-06
Explore housing policy implementation in the Portland Metro Area


danmelton:
# Story Card Request
<img width="554" alt="housing_card_wireframes2" src="https://user-images.githubusercontent.com/99731/39678325-3646e550-513f-11e8-9d57-4e3b77b7981b.png">

**Project: Housing Affordability**

**Card Title: Housing Inventory**

## Description / Wireframe

This card shows the housing inventory policies by governments who have implemented them. Each circle represents a government. When you hover over a government, all the dots for the government go blue. If you hover over a column, the policy.

https://www.figma.com/proto/Sx4NhVdMgKdKKjPV6IEWIX7i/Housing-Card-Wireframes?node-id=202%3A197&scaling=contain

## Data Details (APIs)

https://docs.google.com/spreadsheets/d/1nc_luva5sNZI2X5bIgjg_jPhErbVspJ933AFAnNPYRw/edit#gid=122423246

### How did we calculate this
We interviewed local governments and collected information on the housing policies.

### See the source
>Link to API / data documentation on GitHub
>Link to original source data

jaronheard:
Thinking about the front-end next steps for this, I was wondering how you all would feel about a sortable grid for this? 

Suspend your disbelief and imagine something like this that wasn't horribly ugly, and was sortable, and had the same sort of hovering interaction you imagined.
<img width="588" alt="image" src="https://user-images.githubusercontent.com/7065695/40148785-c7868c72-5924-11e8-92cf-0a454bbd911b.png">

jaronheard:
I think that a sortable grid has a few advantages:
- It's easier to make comparisons
- It's sortable

Either way, we'll want to put some design resources on this component in particular, which I think will be the next step from a design perspective.



I think from a design perspective - we have a data grid component that is in development, and already needs design work.

adpeters:
API calls are similar to #101.
Policy API Sample Filtering by Category: 
http://45.55.145.111:8000/api/policies/?category=regulatory-tools
```
{
    "count": 6,
    "next": null,
    "previous": null,
    "results": [
        {
            "policy_id": "MIZ",
            "policy_type": "Mandatory Inclusionary Zoning",
            "description": "Requires a percentage share of new construction to be affordable by people with low to moderate incomes.",
            "category": "Regulatory Tools",
            "link1": "https://www.huduser.gov/portal/periodicals/em/spring13/highlight3.html",
[...]

PR #391, opened 2018-09-18
Card/ Explore Housing Policy in the Portland Area

jaronheard:
This implements #106, with a simplified design that I worked on with Cat. Some interesting information hierarchy challenges in this one.

image

### DIFF
--- a/packages/2018-housing-affordability/src/components/ExploreHousingPolicyImplementation/PolicyText.js
+++ b/packages/2018-housing-affordability/src/components/ExploreHousingPolicyImplementation/PolicyText.js
@@ -26,9 +26,6 @@ const policyCircle = css`
   padding: 10px;
 `;
 
-const policyText = css`
-`;
-
 const legendScale = [6, 8, 9];
 
 function PolicyText({ onClick, data, selected }) {
@@ -40,7 +37,7 @@ function PolicyText({ onClick, data, selected }) {
           index={legendScale.findIndex(x => data.governments <= x)}
         />
       </div>
-      <div className={policyText}>
+      <div>
       {selected ? (
         <div>
         <h2 className={policyLink}>


DingoEatingFuzz:
Got some questions.

**1. What do the legend markers mean?**

The idea of a legend is great and really helpful for skimming through the list, but I have no idea what low/medium/high means in this context. Is it a measure of commonality? If so, I would use words like rare/uncommon/common. Is it a measure of complexity to implement or enforce? If so, I would use words like simple/typical/difficult. 

I would also add a sentence explanation because even more specific word choice in the legend leaves room for (mis)interpretation.

**2. Do we expect users to be generally interested in policy, or their personal policies?**

I recall some of the original designs optimizing for the "what are the policies in _my_ municipality?" I know this design still has that information, but it is buried. 

**3. Can we do something beyond links for the headings?**

This strikes me as a typical accordion display, but it is masquerading as a list. I think if the clicking items had a bottom border instead of a text decoration it would go a long way towards giving the information structure.

**4. What do you think about adding a caveats/addendums subsection?**

Right now the caveats for each implementing government is listed inline with the list of governments. This makes it even harder to skim the list of implementing governments. If we broke it into two sections, then the list of governments could be quickly scanned and the details for any government with caveats could be read below. Like:

--------------------------

**Implementing governments**

  - City of Beaverton
  - City of Hillsboro
  - Multnomah County
  - City of Portland
  - State of Oregon
  - City of Tigard
  - Washington County

**Caveats by government**

  - City of Hillsboro
    - None. Currently exploring nonprofit-based options. No MULTE.
  - City of Portland
    - Multiple-Unit Limited Tax Exemption Program: Program provides a ten-year property tax exemption on affordable residential housing developments with 20 or more units.
      - City of Portland: Multiple-Unit Limited Tax Exemption (MULTE) Overview ()
  - State of Oregon
    - Covered by Vertical Housing Zones and various grant programs.
  - City of Tigard
    - Nonprofit Corporation Low Income Housing Tax Exemption: Properties approved for the exemption can utilize tax savings for operating costs, and save developer fees for future housing projects.
      - Tigard Affordable Housing Strategies 2002-2016 ()
  - Washington County
    - Multifamily Tax Exemption: Tax exemption for nonprofit multifamily developers that rent housing at 50% AMI. Can get a property tax exemption.
      - Washington County Housing Services ()

-----------------------------

Aside from that bike-shedding, great job tackling this hard info problem. It's always nice seeing new patterns being added to the platform.

jaronheard:
1. What do the legend markers mean?
Right now, it doesn't represent something that is especially meaningful, as it is counting the individual policies - not the implementing governments. I think what is potentially meaningful is percentage of governments implementing the policy. I'm thinking that using the common / uncommon / rare (mythic rare?) structure will work.

2. Do we expect users to be generally interested in policy, or their personal policies?
This design is optimized for organization by policy - with the implicit expectation that users will be interested in policy. I would be interested in exploring how to incorporate that extra dimension, but I don't have a great concept for how it could work.

3. Can we do something beyond links for the headings?
I like the idea of using a bottom border, especially with the removal of link underlining per the style guide and #392. I'll try a few things.

4. Do you think it would help to add a count of implementing governments to each heading?
I think that could be a nice addition. Honestly, something close to this GitHub design, with colored icon, large text for policy name, and smaller text for governments implementing, and some separation between each heading could work nicely.
image

5. What do you think about adding a caveats/addendums subsection?
I love it. Big improvement.

jaronheard:
I did a redesign of this based on @DingoEatingFuzz's comments.

Updated design:
image

Expanded:
image

Feeling pretty good about the new design, but not sure that what I'm showing in terms of common/uncommon/rare implementation is the best possible. I also think the code could use another look.

packages/2018-housing-affordability/src/components/ExploreHousingPolicyImplementation/index.js
```diff
@@ -1,43 +1,179 @@
...
-          <br />
-          <br />
-          <a href="https://docs.google.com/spreadsheets/d/1SVs9AQkw5guxEIuV1eGASKCg5miWhPn3UoZrALQpXLI/edit?usp=sharing">Explore Findings</a></p>
-
+        </p>
+        <p>
+          In the spring and summer of 2018,{' '}
+          <a href="http://www.hackoregon.org/">Hack Oregon</a> volunteers
+          surveyed local governments to understand what policies were being
+          implemented. Taking inspiration from the{' '}
```
DingoEatingFuzz: What a cool jsx quirk.

packages/2018-housing-affordability/src/state/housing-policy/selectors.js
```diff
@@ -0,0 +1,96 @@
...
+  (policies, programs, selectedPolicy) => {
+    if (!policies || !programs || !selectedPolicy) return;
+
+    const implementingGovernments = programs.filter(
+      program => program.policy === selectedPolicy
+    );
+
+    const selectedPolicyData = policies.filter(
+      policy => policy.policy_id === selectedPolicy
+    );
```
DingoEatingFuzz: You could use `policies.find` here instead. Then you won't have to do all the `selectedPolicyData[0]` stuff below.

packages/2018-housing-affordability/src/components/ExploreHousingPolicyImplementation/index.js
jaronheard: Been using prettify and didn't actually realize what was going on at first - I guess it's just so that you can see the whitespace when you're reading the code.

jaronheard:
@DingoEatingFuzz Thanks for the comments. I refactored the confusing code and incorporated the other suggestions, so I think this should be ready to go.

packages/2018-housing-affordability/src/components/ExploreHousingPolicyImplementation/index.js
DingoEatingFuzz: It's more nefarious than that. [JSX strips all newlines between elements](https://github.com/facebook/react/issues/1643). This is contrary to HTML which treats newlines as whitespace characters. So Prettier must be protecting you from losing the expected space between `the` and the following `a` element.

PR #393, opened 2018-09-18
Mobile fixes

jaronheard:
Some quick fixes for some wonky mobile overlaps


jaronheard:
@DingoEatingFuzz I think that the multi-column layout is something that should make it's way upstream. I'll think about doing that the next time that it comes up. 

The map fix I think is something that is specific to this card, because the square map specified in the design isn't necessarily a design pattern that I see continuing on.

PR #400, opened 2018-10-25
Improve collection page layout for mobile

jaronheard:
This PR improves the page layout for mobile by doing a few things:

- Fixes an issue with the Nav that caused white-space to the right on smaller phones
- Decreases the size of collection titles to prevent long words from causing overlaps
- Increases card width to almost full-screen width on mobile
- Decreases horizontal padding within CivicStoryCard on mobile

I actually think a full mobile card redesign is in order, in order to fully utilize limited horizontal space, but this is a big improvement, especially for the data visualizations.

Any feedback appreciated!

The most dramatic improvement is at the smallest sizes - see iPhone SE example below:

Before                                       |  After
:-------------------------:|:-------------------------:
image | image
image | image
image | image


packages/component-library/src/Hero/CollectionHero.js
```diff
@@ -6,16 +6,27 @@ import { css } from 'emotion';
...
+    font-size: 36px;
+  }
+`;
+
 const CollectionHero = ({ heroTitle, heroSubtitle, mainProjectColor, teamTitle }) => (
   <Hero mainProjectColor={mainProjectColor}>
     <div>
       <span className={teamTitleStyle}>{teamTitle}</span>
-      <h1 className={'Title'}>{heroTitle}</h1>
+      <h1 className={titleStyle}>{heroTitle}</h1>
```
jaronheard: This was the only place in 2018 code where we were using 'Title' from the global css file - so changed it here rather than changing it globally

PR #417, opened 2019-02-20
RFC: Add Automatic Code Formatting Using Prettier

jaronheard:
## Overview
**What:** Add [Prettier](https://prettier.io/) to our toolchain for automatic code formatting
**Why:** Removing having to worry about code formatting will to focus on making things that work. It makes things easier for new contributors. It automatically resolves any future individual "code style" discussions. Prettier has [more reasons](https://prettier.io/docs/en/why-prettier.html).
**How:**  Example implementation included in this PR, but TBD
**When:** Discussion by February 28, implementation (if applicable) by March 14

This RFC shows a sample implementation of adding [Prettier](https://prettier.io/), and running Prettier on the component library package to show an example of the code formatting that would be automatically applied.

Removing having to worry about code style allows us to focus on making things that work. It makes things easier for new contributors. It automatically resolves any future individual "code style" discussions.

### How do we want Prettier to be implemented?
The example implementation in this PR uses a precommit hook with [lint-staged](https://github.com/okonet/lint-staged) and [husky](https://github.com/typicode/husky), as described in the [Prettier documentation](https://prettier.io/docs/en/precommit.html). There are other options: precommit hooks could be implemented differently, or alternatively, Prettier could be run "on-change".
### What is the developer workflow for this implementation?
In short, code in your preferred formatting, and your code will be automagically formatted on commit✨. Technically, when you make a commit, Prettier will be run as a pre-commit hook on all staged files matching (js,js,css,md).

### Prettier configuration
Prettier has configuration options. I've used the default options, but if we want something other than the default, now's would be the time.

### In what ways should our ESLint configuration be adjusted?
My observation is that our current ESLint configuration has in practice been ignoring ESLint complaints, as many of them are related to code styling.
### What are compelling reasons *not* to do this?
It adds additional complexity to the toolchain, and makes making a commit slower.

## Next steps
- [x] Feedback: Do we want to do this? How? (By February 28)
- [x] Implementation - Implement desired Prettier implementation
- [x] Implementation - Run Prettier on entire project
- [ ] Implementation - Add "Code Style" section to documentation


jaronheard:
I received some verbal feedback from @DingoEatingFuzz, which I'm recording here. In a nutshell:

- He's for implementing Prettier
- Re: implementation of Prettier, he thought that a pre-commit hook using lint-staged and husky was 💯
- He also advocated for a more strict implementation of ESLint, his preferred approach being that not passing ESLint causes Travis to fail tests.

DingoEatingFuzz:
As you have mentioned, I am a huge +1 to this. My thoughts looking at this PR are "holy diff size". It looks like there is some weird interaction with the giant json files we have in the repo.

Big picture: those json files should go away at some point. Small picture: I don't think we want prettier to touch json files.

jaronheard:
I modified the Prettier settings to exclude JSON files and reran Prettier on the whole project.

I think that this is ready to be merged pending review, and the ESLint enforcement and code style documentation should be worked on in a separate PR.

As the Hack Oregon season starts, people are starting to poke around in this repo, and think this will make that experience much more pleasant.

Issue #427, opened 2019-03-16
Sandbox story doesn't show foundation or slide layers


mendozaline:
When you open the [sandbox story](https://hackoregon.github.io/civic/?selectedKind=CIVIC%20Platform%20Components%2F%2FSandbox&selectedStory=Sandbox&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs) in the Hack Oregon storybook, the foundation and slide layers don't appear.
![Screenshot_2019-03-16 Storybook](https://user-images.githubusercontent.com/9361258/54480930-cafd5f00-47eb-11e9-9006-efefbb0d017b.png)

PR #428, opened 2019-03-16
Fix sandbox story

mendozaline:
This address #427. 

The constant.js file just needed to be updated.
Screenshot_2019-03-16 Storybook(2)

I also updated the map style to Hack Oregon light.


mendozaline:
Good suggestions, Jaron. I identified the required props for the Sandbox and added "isRequired" to them. I also removed the mapboxToken and mapboxStyle props.

PR #431, opened 2019-03-18
Remove GeoJSON files from component library

mendozaline:
This is a follow up to #258 and #417. I removed GeoJSON files used in the ScreenGrid Map, ScatterPlot Map, Boundary Map, Path Map, Icon Map, and CivicSandbox Map stories. 

I added a helper component named LoadData that takes an array of urls as a prop and then fetches and passes that data down to it children. So now the map stories are using real data from our APIs and not local GeoJSON files.

I also updated and simplified the stories by removing some redundant knobs.


mendozaline:
I was worried about this too. What if I renamed it something like StorybookJSONLoader instead?

jaronheard:
@mendozaline What about DemoJSONLoader instead? Just in case we ever migrate from Storybook.

PR #442, opened 2019-03-29
Upgrade 2017 Emergency Response Components

jaronheard:
This upgrades the 2017 Emergency Response components to use our latest components, quick and dirty like, so we can get Leaflet and Recharts out of component library to finish this dependency update. Screenshot below:

localhost_3000_


packages/component-library/src/utils/civicFormat.test.js
```diff
@@ -19,6 +19,7 @@ describe('civicFormat', () => {
 
   it('should format dollars correctly', () => {
     expect(civicFormat.dollars(12)).to.eql('$12');
-    expect(civicFormat.dollars(230.5)).to.eql('$230.5');
+    expect(civicFormat.dollars(230.5)).to.eql('$231');
+    expect(civicFormat.dollars(7300000)).to.eql('$7.3 million');
```
DingoEatingFuzz: The removed test made me think about how that's not how money is presented. It would be nice to have a test case for `230.5 -> $230.50`, but no pressure to add that as part of this.

packages/emergency-response/src/components/FmaMap/index.js
```diff
@@ -1,104 +1,61 @@
...
+                    </BaseMap>
+                )}
+                {fmaPanelData && !isEmpty(fmaPanelData) ? (
+                    <MapPanel />
+                ) : (
+                    <h6>Click on an FMA to view some statistics</h6>
+                )}
+            </div>
+        )
+    }
```
DingoEatingFuzz: This is four space indented and I hate it 💃 

Can't wait for the impending prettier + eslint PR after the upgrade one. My vote is still for two space indentation, but maybe we should take it to a poll.

jaronheard: I also can't wait.

Issue #443, opened 2019-03-29
Sandbox - Data Selection Side Menu


mendozaline:
This issue incorporates the following changes to the Civic Sandbox UI:
- [x] Create a side panel for all data selectors
- [x] Move Data Collection dropdown menu into side panel 
- [x] Move Base Layer dropdown menu into side panel
- [x] Include Base Layer color legend 
- [x] Move Data Overlays selector into side panel
- [x] Show all slide options for a Data Collection
- [x] Embiggen Base Map

![sketch-00](https://user-images.githubusercontent.com/9361258/55264752-a9ac6200-5232-11e9-87c8-285cc13fe5ed.png)

PR #445, opened 2019-04-02
Civic Sandbox Improvements

mendozaline:
This addresses #443. 

The main changes involve how the user selects Data Collections, Base Layers, and Data Overlays. Instead of dropdown menus along the top, the new design has everything in a side panel on the right. 

These changes also allow the BaseMap to be larger. This is done with a new `useContainerHeight` prop that works the same way as how the Base Map's width was made responsive with React Dimensions. 

Full view:
sandbox-00

Mobile view:
00-sandbox

Other changes include:
- Swapped `componentWillReceiveProps` in the Sandbox with `componentDidMount` and `componentDidUpdate`, since `componentWillReceiveProps` is [no longer recommended](https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops). This should also make it easier to use hooks once we upgrade to React 16
- Removed the dropdown multi-selector and replaced it with toggle switches to select slides
- The background color of each slide's title corresponds to the color of that slide on the map. So this now also functions as a legend
- Color scales for most maps are now being created dynamically inside the CivicSandboxMap component using d3 scales
- The Base Layer's color scale legend has been moved from the dashboard to the side menu. It now also includes tick values
- Added an optional placeholder prop for the Dropdown component


packages/component-library/src/CivicSandboxMap/CivicSandboxMap.js
```diff
@@ -11,6 +11,7 @@ import DeckGL, {
   ScreenGridLayer,
 } from 'deck.gl';
 import { css } from 'emotion';
+import * as d3 from 'd3';
```
jaronheard: I'm trying to move away from importing * as d3 to importing individual modules from d3 so we get the benefits of that modularization. We're doing it some other places, so maybe it makes sense to do this as a separate issue

packages/component-library/src/Sandbox/Sandbox.js
```diff
@@ -105,15 +70,19 @@ const Sandbox = ({
...
+          updatePackage={updatePackage}
+          updateFoundation={updateFoundation}
+          foundationMapProps={foundationMapProps}
         />
       </div>
-      <div>
+      <div className={baseMapWrapper}>
         <BaseMap
-          mapboxStyle={mapboxStyle}
+          mapboxStyle={"mapbox://styles/mapbox/dark-v9"}
```
jaronheard: I think it makes sense to pull this style into BaseMap so that our different map styles are tracked in one place. Maybe we could make our own "civicMapStyle" prop, with light and dark as options.

packages/component-library/src/Sandbox/SandboxMapLegend.js
```diff
@@ -0,0 +1,114 @@
+import React from 'react';
+import * as d3 from 'd3';
```
jaronheard: See previous * as d3 comment

packages/component-library/src/Sandbox/SandboxMapLegend.js
```diff
@@ -0,0 +1,114 @@
...
+  const bins = mapProps.scaleType === 'ordinal' || mapProps.scaleType === 'threshold'
+    ? colorScale
+    : colorScale.range().map(d => colorScale.invertExtent(d));
+
+  const ticks = mapProps.scaleType === 'ordinal' || mapProps.scaleType === 'threshold'
+    ? bins
+    : bins.reduce((a, c, i, arr) => c[1] ? [...a, c[1]] : [...a, ''], []);
+
+  const thousandsFormat = d3.format('.3s');
+  const percentFormat = d3.format('.1%');
```
jaronheard: Do either of these make sense to pull into civicFormat?

mendozaline:
Thanks for the great suggestions, Jaron. I'll create some issues to address these items.

PR #450, opened 2019-04-03
Use Victory and Mapbox in Transportation 2017 project

DingoEatingFuzz:
No more Leaflet, no more Recharts.


DingoEatingFuzz:
Um, possibly? It seems unlikely that someone would ever be sending props to the basemap without wanting those props to take effect though.

jaronheard:
Yeah, think you're right. Feel free to merge at your leisure 🛳🏖

Issue #462, opened 2019-04-10
Add civicMapStyle prop to BaseMap component


mendozaline:
Per Jaron in [#445](https://github.com/hackoregon/civic/pull/445#discussion_r273321070):
> I think it makes sense to pull this style into BaseMap so that our different map styles are tracked in one place. Maybe we could make our own "civicMapStyle" prop, with light and dark as options.

- [x] Add `civicMapStyle` prop to BaseMap component
- [x] Should have `dark` and `light` options
- Light option: `"mapbox://styles/hackoregon/cjiazbo185eib2srytwzleplg"`
- Dark option: `"mapbox://styles/mapbox/dark-v9"`
- [x] Default should be light option
- [ ] Update BaseMap story to include an example with the `civicMapStyle` prop

mendozaline:
Addressed by #482.

PR #482, opened 2019-04-22
Add SandboxBaseMapSelector component 

paigewilliams:
per issue #462 

The main changes include:

- Creating the SandboxBaseMapSelector component that is a form consisting of radio buttions to select either a light or dark basemap option.

- Converting the Sandbox component into a stateful component with state values of BaseMapStyle and BaseMapStyleUrl. These values are passed to the SandBoxBaseMapSelector as props to check light or dark in the form.

Screen Shot 2019-04-21 at 7 42 13 PM
Screen Shot 2019-04-21 at 7 41 56 PM

I decided to make this a stateful component because it is just UI state. If converting this to a functional component and using Redux is preferred, I am happy to implement that. I also thought about making this a toggle like the toggles in the sidebar. 

Looking forward to feedback 💃


jaronheard:
Cool stuff @paigewilliams! Excited for your first PR on CIVIC!!!  🙌🚀🙌

I think a map style selector is a cool feature for the Sandbox, and the implementation looks clean! ✨

I'm going to defer to @mendozaline on how this fits into the Sandbox UI and whether this should be stateless in some way -- using state makes sense to me, as you stated it's component specific UI state.

In order to close out #462, I'd like to extract out the BaseMap styles to the BaseMap component -- there are more details on the desired implementation in that issue. The idea behind the "civicMapStyle" prop to the BaseMap component and storing the "light" and "dark" mapbox urls there is rather than having hardcoded mapbox urls in different places, they'll all just be in one place.

So you could simplify your code by removing the baseMapStyleUrl from your code and make the change below:
```
<BaseMap
   mapboxStyle={this.state.baseMapStyleUrl}
...
// would change to
<BaseMap
   civicMapStyle={this.state.baseMapStyle}
...
```

packages/component-library/src/Sandbox/SandboxBaseMapSelector.js
```diff
@@ -0,0 +1,43 @@
...
+  return (
+    <>
+      <div className={baseMapMenu}>
+        <input
+          checked={baseMapStyle === "light"}
+          type="radio"
+          value="light"
+          onChange={onBaseMapStyleChange}
+        />
+        <label for="light">Light</label>
```
mendozaline: Great job taking accessibility into account here. Just note that when using JSX to replace `for` with `htmlFor` in labels and to include an `id` with each input. 

I found this page from the React documentation useful when I had to make similar changes to the SandboxToggleSwitch component: https://reactjs.org/docs/accessibility.html#accessible-forms

packages/component-library/src/Sandbox/SandboxBaseMapSelector.js
```diff
@@ -0,0 +1,43 @@
...
+  padding: 10px;
+  bottom: 0;
+  right: 0;
+  z-index: 2;
+  font-family: "Roboto Condensed", sans-serif;
+`);
+
+const SandboxBaseMapSelector = ({ onBaseMapStyleChange, baseMapStyle }) => {
+  return (
+    <>
```
mendozaline: Might consider using `<React.Fragment></React.Fragment>` here to get rid of the warning.

paigewilliams:
@jaronheard @mendozaline Thanks so much for the awesome feedback!! 💯 💯 

I think moving the `SandboxBaseMapSelector` to the `SandboxDrawer` makes sense. I also think that making it a functional component and using the `useState` hook would clean it up a bit.

Just so I am understanding this correcty, would we like to move the basemap urls to local state in the `BaseMap` component and mutate it there?

jaronheard:
@paigewilliams you can just keep the BaseMap component functional (adding a prop for civicMapStyle) and hold the state in the SandboxBaseMapSelector — but do move the urls

I think #462 has a pretty detailed description of how civicMapStyle should work

Jaron Heard
Creative Director
Civic Software Foundation

+1-971-998-7180

paigewilliams:
@jaronheard @mendozaline I have updated the PR with your requests. My changes include the following: 

- [x] Use `<React.Fragment>` in the SandboxBaseMapSelector
- [x] Use `htmlFor` for the radio button labels
- [x] Move the `BaseMapSelector` to the `SandboxDrawer component` 
- [x] Refactor the `Sandbox` component to a functional component. I used the `useState` hook to set the `baseMapStyle` to either light or dark. 
- [x] Extract out the BaseMap styles to the BaseMap component. The two urls are currently stored in constants, that are triggered in the `componentWillReceiveProps` lifecycle method. 

Screen Shot 2019-04-28 at 7 48 10 PM

mendozaline:
The changes look great, Paige. Thanks for taking the time to make them.

Assuming that @jaronheard doesn’t have any suggestions or changes, everything else looks fine to merge.

Issue #464, opened 2019-04-10
Sandbox - Civic Sandbox Dashboard Improvements


mendozaline:
This issue incorporates the following changes to the Civic Sandbox Dashboard component:

- [x] Reposition to the bottom left 
- [x] Make mobile friendly 
- [x] Make dashboard collapsible 
- [x] Improvements to data visualization components 
- [x] Various styling changes

mendozaline:
Addressed by #483.

PR #404, opened 2018-11-09
RFC - Civic Sandbox Cards(!)

jaronheard:
This PR sets up an initial front-end implementation for Sandbox cards, along with implementing a demonstration card within the Disaster Resilience story collection.

In addition, it makes the following changes related to implementing Sandbox cards:
- Improvements to CivicSandboxDashboard component
- Extracted CivicWatermark component for use in CivicSandboxDashboard and CivicStoryCard, which adjusts size by default using ReactDimensions, but can be passed a "small" prop.

There are a few areas where I would love feedback. @mendozaline @DingoEatingFuzz @nicrocs 
- Design and code feedback for initial components
- Advice for implementation for the following items:
  - How to facilitate best reuse of existing Sandbox code for data loading? Perhaps it makes sense to [incorporate the work that @AraOshin did on the back-end](https://github.com/AraOshin/civic-sandbox-backend) prior to building something new here.
  - How to best link into the full Sandbox view in the future. This may be pretty involved, given the existing Sandbox design

Sample images below:

Sandbox card in collection, desktop:
image

Sandbox card in collection, mobile
image

Showing updated Sandbox dashboard in full Sandbox view:
image

PR #483, opened 2019-04-22
Civic Sandbox Dashboard Improvements

mendozaline:
This addresses #464.

The main changes to the Civic Sandbox Dashboard include: 
- Makes the dashboard collapsible 
- Highlights the selected polygon
- Removes the comparison bars option for now. It can be added back when we have accurate data to display instead of the placeholders values currently used
- Transitions the Civic Sandbox Dashboard component from a class to functional component thanks to the `useState` hook 

Other changes:
- Removes the Base Layer's color legend from the dashboard, since it has been moved to the side menu
- Fixes a bug that resets the map's zoom level when you activate the tooltip or open the side menu by adding an `updateViewport` prop to the BaseMap. When set to `false`, the `componentWillReceiveProps` lifecycle is bypassed. `updateViewport` is set to `true` by default.
- Ensures the accessibility of the Sandbox Toggle Switch component by adding `htmlFor` to the labels and an `id` to the inputs
- Update component `propTypes` to use `shape({})` and add more useful information

dashboard-closed
dashboard-open


packages/civic-sandbox/src/state/sandbox/selectors.js
```diff
@@ -257,7 +171,7 @@ export const getSelectedFoundationDatum = createSelector(
       attrs.secondary.field
     );
 
-    if (!priamryFieldMatch && !secondaryFieldMatch) return;
+    if (!priamryFieldMatch && !secondaryFieldMatch) return [];
```
jaronheard: Looks like primary is misspelled here

mendozaline:
@jaronheard Yes, I agree that the dashboard could use the improvements made in #404. It would be great to have those changes included in the next iteration.

PR #488, opened 2019-04-23
Updates to Design System for Storybook v5

bradfordjanice:
Reorganized story index, customize addons panel, and added a vision statement to Accessibility story.


jaronheard:
Note: c552a69 passes, but it has unresolved merge conflicts in the AccessibilityGuidelinesStyle.story.js -- we should make sure that we don't get a green light from Travis on the component library on things that will prevent Storybook from building

packages/component-library/stories/styleGuideStories/AccessibilityGuidelinesStyle.story.js
```diff
@@ -27,6 +32,7 @@ const AccessibilityGuidelines = () => (
       Learning disabilities, distractibility, and inability to remember or focus
       on large amounts of information.
     </p>
+<<<<<<< HEAD
```
jaronheard: Looks like there's still a merge conflict to resolve in this file.

PR #493, opened 2019-04-27
Update 2017 housing pie chart to useLegend

jaronheard:
Before:
image

After:
image

Additionally, was able to unprotect labels, and fixed a bug in useLegend implementation.


packages/component-library/src/PieChart/PieChart.js
```diff
@@ -29,7 +29,7 @@ const PieChart = props => {
   const y = getOrElse(dataValue, "y");
   const startAngle = halfDoughnut ? -90 : 0;
   const endAngle = halfDoughnut ? 90 : 360;
-  const legendLabels = data.map(value => ({ name: value.x }));
+  const legendLabels = data.map(value => ({ name: value[x] }));
```
BJHunnicutt: Woops, good call!

BJHunnicutt: Woops, good call!

Issue #496, opened 2019-04-30
Add viewport add-on to Storybook


jaronheard:
### What

Add the [viewport add on](https://github.com/storybooks/storybook/tree/master/addons/viewport) to Storybook.

### Why

Developers and designers should be able to see what components look like in the context of different devices.

### When

As soon as possible.

jaronheard:
#497 addressed

PR #497, opened 2019-04-30
Update Storybook to 5.0.11, add Viewport add-on

jaronheard:
Addresses #496, and updates Storybook to 5.0.11

Doesn't change the default:
image

But there when you want it:
image


bradfordjanice:
This is a great add-on; it will be very helpful for designers!

I see there is a conflict that needs to be resolved...

PR #506, opened 2019-05-03
Design system/logo

bradfordjanice:
Used Storybook theming to replace the Storybook logo with the CIVIC logo on our Storybook site.  See Storybook's instructions here:
https://github.com/storybooks/storybook/blob/master/docs/src/pages/configurations/theming/index.md

It should show the square logo followed by "CIVIC Platform", however it doesn't appear in Chrome or Firefox, and in Edge only the logo appears.


jaronheard:
Note, this branch needs troubleshooting and is ready for review

mendozaline:
I think I figured out the problem here:

1. The reason the logo doesn't appear is because [according to this issue](https://github.com/storybooks/storybook/issues/5878#issuecomment-470475463) the svg being imported needs to have a `width` and `height` attribute set. If you go into the file that contains the Civic logo and add a `width` and `height` attribute to the svg element and set both to 50, the Civic logo now appears.

2. However, the `brandTitle` of "CIVIC Platform" still doesn't appear next to it. This is because `brandTitle` is used to set the `alt` attribute of the logo image, not as a title that appears next to the logo image. [According to this issue](https://github.com/storybooks/storybook/issues/5866#issuecomment-473344654), it's not obvious that the logo the default storybook theme uses is really 1 svg with a path element for the "S" logo and another path element for the text "Storybook" next to it. The Storybook documentation doesn't really make clear what the purpose of `brandTitle` is.

So it looks like we can use the Civic logo svg and add a `height` and `width` attribute to it. Or create a new svg or png with the Civic logo and the text "CIVIC Platform" next to it.

bradfordjanice:
Thank you very much for your research on this issue!  Your solution was right on and I've updated the logo SVG file and the CIVIC theme.

PR #508, opened 2019-05-03
Update/storybook button knobs

bradfordjanice:
A work in progress...

Button Story:
* Renamed and created 3 button stories for different development goals.
* Added Notes - read notes to understand the differences in the 3 stories.
* Added knobs to each story based on how the developer will want to use the button in their design; knobs were added only for those props that can be edited within the design goals of the CIVIC platform.

Please review:
* Look at the presentation in these 3 browsers:  Chrome, Firefox, Edge
* Do the 3 stories and knobs for each support the way the CIVIC platform is meant to be used and stay within the design goals of the platform?
* Are the Notes complete enough?  Have too much information?  
* The Story code in the addon panel... the goal is to write the story so the developer will be able to see which props they use when they want to use that "version" of a component.  Do the button stories do this?

Still to do:
* Add a knob for changing the label on the non-text label button
* Get the knob for enable/disable button working 
* A story for a multi-select dropdown list

Issues:
* The order of stories isn't correct in Firefox.
* Additional knobs are sometimes shown (could this be a bug in Storybook?):  Data, Data key, Data values, Data key labels
* Notes are not using the correct font family; I will add an issue to update the Storybook theme.
* The red color (CIVIC color theme) used in the button fails accessibility test; I will add an issue to update the color theme to meet accessibility requirements.

### DIFF
--- a/packages/component-library/stories/Button.story.js
+++ b/packages/component-library/stories/Button.story.js
@@ -8,6 +8,7 @@ import {
   text,
   color,
   boolean,
+  optionsKnob as options,
   select
 } from "@storybook/addon-knobs";
 import notes from "./button.markdown.md";
@@ -51,7 +52,16 @@ export default () =>
         "all .2s ease-in-out"
       );
       //const commonTransition = text("Transition", "all .2s ease-in-out");
-      const disabled = boolean("Disabled", false);
+      //const disabled = boolean("Disabled", false);
+      const valuesObj = {
+        Enabled: "",
+        Disabled: "disabled"
+      };
+      const optionsObj = {
+        display: "inline-radio"
+      };
+      const disabled = options("State", valuesObj, "", optionsObj);
+      //console.log(disabled);
       return (
         <Button
           onClick={action("clicked")}
--- a/packages/component-library/stories/Button.story.js
+++ b/packages/component-library/stories/Button.story.js
@@ -3,20 +3,23 @@ import React from "react";
 import { storiesOf } from "@storybook/react";
 import { action } from "@storybook/addon-actions";
 import { checkA11y } from "@storybook/addon-a11y";
-import { withKnobs, text } from "@storybook/addon-knobs";
+import {
+  withKnobs,
+  text,
+  color,
+  boolean,
+  select
+} from "@storybook/addon-knobs";
+import notes from "./button.markdown.md";
 import { Button } from "../src";
 import { storybookStyles } from "./storyStyles.js";
-
-const displayName = Button.displayName || "Button";
 
 const description = `
   This is some basic usage with the button with providing a label to show the text.
   Clicking should trigger an action.`;
 
-const simpleButtonLabel = "Hello button";
-
 export default () =>
-  storiesOf("Component Lib|Basic UI Components/Button", module)
+  storiesOf("Component Lib|Basic Inputs/Button", module)
     .addDecorator(withKnobs)
     .addDecorator(checkA11y)
     .addDecorator(story => (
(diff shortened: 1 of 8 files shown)


packages/component-library/stories/Button.story.js
```diff
@@ -3,34 +3,103 @@ import React from "react";
...
 import { checkA11y } from "@storybook/addon-a11y";
-import { withKnobs, text } from "@storybook/addon-knobs";
+import {
+  withKnobs,
+  text,
+  color,
+  optionsKnob as options,
+  select
+} from "@storybook/addon-knobs";
+import notes from "./button.markdown.md";
```
jaronheard: Recommend renaming `button.markdown.md` to `button.md` or `button.notes.md`

packages/component-library/stories/Button.story.js
```diff
@@ -3,34 +3,103 @@ import React from "react";
...
-      <Button onClick={action("clicked")}>
-        <span role="img" aria-label="So cool">
-          😀 😎 👍 💯
-        </span>
-      </Button>
-    ));
+    .add(
+      "Standard",
+      () => {
+        const label = text("Label", "Hello CIVIC");
```
jaronheard: Recommend pulling the knobs outside of the function block body so they can be reused across stories.

```
import { storybookStyles } from "./storyStyles.js";

// define knobs here so they can be reused across stories
const label = text("Label", "Hello CIVIC");
…

export default () =>
…

    .add(
      "Standard",
// can use concise syntax here
      () => (<Button onClick={action("clicked")}>{label}</Button>),
      { notes }
    )
```

packages/component-library/stories/Button.story.js
```diff
@@ -3,34 +3,103 @@ import React from "react";
...
+        //const disabled = boolean("Disabled", false);
+        const valuesObj = {
+          Enabled: "",
+          Disabled: "disabled"
+        };
+        const optionsObj = {
+          display: "inline-radio"
+        };
+        const disabled = options("State", valuesObj, "", optionsObj);
+        //console.log(disabled);
```
jaronheard: Per above, recommend pulling the knobs outside of the function block body so they can be reused across stories.

packages/component-library/stories/Button.story.js
```diff
@@ -3,34 +3,103 @@ import React from "react";
...
+          Standard: "all .2s ease-in-out",
+          Slow: "all .6s ease-in-out",
+          Fast: "all .1s ease-in-out"
+        };
+        const transition = select(
+          "Transition",
+          transitionOptions,
+          "all .2s ease-in-out"
+        );
+        const ariaLabel = text("aria-label", "So cool");
```
jaronheard: Per above, recommend pulling the knobs outside of the function block body so they can be reused across stories.

packages/component-library/stories/button.markdown.md
```diff
@@ -0,0 +1,36 @@
+# Button Component
+
```
jaronheard: Per our in person discussion, recommend continuing to explore options here that meet the following criteria:

- Valuable to design system users
- Simple to keep in sync with current state of code
- Future compatibility / adaptation to Storybook docs

packages/component-library/stories/Button.story.js
jaronheard: Note, in case anyone else is following along -- what I suggested doesn't work. Knobs need to be defined withing the function block as part of .add

packages/component-library/stories/Button.story.js
jaronheard: As notes above, doesn't work

packages/component-library/stories/Button.story.js
jaronheard: As notes above, doesn't work

packages/component-library/stories/button.markdown.md
bradfordjanice: All of the above requested changes have been made.

packages/component-library/stories/button.notes.md
```diff
@@ -0,0 +1,22 @@
...
+- Text label
+
+## Custom Non-text Label
+
+Use the Non-text Label (e.g., emojis or symbols) button if the label isn't readable by accesibile technologies, e.g. it is an emoji or other image. Use the aria-label attribute to give the button an accessible name. See https://www.w3.org/TR/WCAG20-TECHS/ARIA14.html for more information about the aria-label.
+
+This story allows the developer to edit these attributes of the Standard button:
+
+- Non-text label
+- aria-label
```
jaronheard: I took a stab at editing this a little bit more -- let me know what you think

# Button Component

## Standard

The Standard story shows uses the standard usage of the Button for the CIVIC platform. Button includes standard styling.

## Custom

The Custom story shows all possible properties that can be passed to Button component for customization.

- Color theme
- Transition
- Text label

## Example: Non-text Label

This Non-text Label example shows how to use Button if the label isn't readable by accessible technologies, e.g. it is an emoji or other image. Use the aria-label attribute to give the button an accessible name. [More information about aria-label](https://www.w3.org/TR/WCAG20-TECHS/ARIA14.html).

This story allows the developer to edit these attributes of the Standard button:

- Non-text label
- aria-label

packages/component-library/stories/Button.story.js
```diff
@@ -3,34 +3,69 @@ import React from "react";
...
 import { storybookStyles } from "./storyStyles";
-
-const displayName = Button.displayName || "Button";
-
-const description = `
-  This is some basic usage with the button with providing a label to show the text.
-  Clicking should trigger an action.`;
-
-const simpleButtonLabel = "Hello button";
+import { file } from "@babel/types";
```
jaronheard: Is this needed? Is this needed in every file where we import notes?

bradfordjanice: No, this is required if you are using the file knob.  I experimented with using it in this story, but missed removing this import in the final design.  It's been removed.

Issue #515, opened 2019-05-09
Improve code quality


jaronheard:
This epic tracks progress for improving code quality through linting and automated formatting:

- [x] Run prettier on project #465 
- [x] Implement prettier hook #467 
- [x] Lint component library #468 
- [x] Lint 2018 package #517  
- [x] Lint 2018-disaster-resilience package #518 
- [x] Lint 2018-transportation-systems package #519
- [x] Lint 2018-housing-affordability package #520 
- [x] Lint 2018-neighborhood development package #521
- [x] Lint 2018-local-elections package #522 
- [x] Implement linting hook #523
- [x] Lint convenience packages #592 

**After Demo Day list**
Add prettier and linting checks to CI #591
Lint 2017 project #593 
Assorted linting follow ups, listed below

**Assorted follow ups**
Enable and fix linting for Disaster Resilience Quiz Card (WhatCanYouDoToPrepare)

PR #517, opened 2019-05-09
Lint 2018 package

jaronheard:
Addresses #515 for 2018 package


Dianna:
@jaronheard, doesn't look like I have the necessary permissions to assign myself as a reviewer. Let it be known that I shall, henceforth, review-ify these changes

packages/2018/src/components/HomePage.js
```diff
@@ -697,10 +702,10 @@ class HomePage extends Component {
                   </p>
                 </div>
                 <div className={brainWrapper}>
-                  <img src={brain} />
+                  <img src={brain} alt="" />
```
Dianna: Not sure what the page looks like, so depending on that an alt may be nice here

packages/2018/src/components/HomePage.js
```diff
@@ -697,10 +702,10 @@ class HomePage extends Component {
...
                 </div>
                 <div className={brainWrapperMobile}>
-                  <img src={brainMobile} />
+                  <img src={brainMobile} alt="" />
```
Dianna: Same about the alt

packages/2018/src/components/HomePage.js
jaronheard: Yeah, these are decorative, so empty alt is intentional! Thanks for thinking about it!

packages/2018/src/components/HomePage.js
jaronheard: Also decorative

jaronheard:
Thanks for the review @Dianna! I've added you to the front-end team, so you have all the appropriate permissions now

PR #521, opened 2019-05-09
Lint 2018-neighborhood-development package

jaronheard:
Addresses #515 for 2018-neighborhood-development package


packages/2018-neighborhood-development/src/components/HumanImpactOfSweepingUrbanCampsites/index.js
```diff
@@ -66,7 +68,7 @@ export class HumanImpactOfSweepingUrbanCampsites extends React.Component {
           <Collapsable.Section hidden>
             <div>
               <p>
-                The Homelessness/Urban Camping Impact Reduction Program's
+                The Homelessness/Urban Camping Impact Reduction Program&apos;s
```
BJHunnicutt: If this won't change again then this probably doesn't matter here, but in the future, turning this into a template string might keep the text more readable. 
So instead of:
`&quot;a&quot;, &quot;list&quot;,  &quot;of&quot;, &quot;quotes&quot;`
 It would just be: 
`{'"a", "list", "of", "quotes"'}`

PR #523, opened 2019-05-10
Implement eslint hook in lint-staged

jaronheard:
This adds eslint to the lint-staged hook currently used to prettier files for #515. It will display linting errors in the console, and not allow you to commit unless you fix them (or manually skip the process).

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,11 +48,12 @@
     }
   },
   "lint-staged": {
-    "*.{js,jsx,json}": [
+    "*.{js,jsx}": [
       "prettier --write",
+      "eslint --fix",
       "git add"
     ],
-    "*.{css,md}": [
+    "*.{css,md,json}": [
       "prettier --write",
       "git add"
     ]
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "start": "node scripts/start",
     "deploy-storybook": "BABEL_ENV=esm storybook-to-ghpages",
     "in": "lerna run --scope",
-    "lint": "lerna run lint",
+    "lint": "lerna run lint --no-bail",
     "lint:fix": "lerna run lint --no-bail -- --fix",
     "storybook": "BABEL_ENV=esm start-storybook -p 6006",
     "test": "lerna run test --no-bail",
--- a/packages/budget/package.json
+++ b/packages/budget/package.json
@@ -29,7 +29,7 @@
     "start:dev": "cross-env NODE_ENV=development node server",
     "start:prod": "cross-env NODE_ENV=production node server",
     "test": "cross-env BABEL_ENV=test mocha --opts ./mocha.opts 'src/**/*.test.js'",
-    "lint": "eslint --fix src"
+    "lint": "eslint src"
   },
   "dependencies": {
     "@babel/polyfill": "^7.0.0",
--- a/packages/2017/package.json
+++ b/packages/2017/package.json
@@ -10,7 +10,7 @@
     "start:prod": "yarn run build && cross-env NODE_ENV=production node server",
     "build": "rimraf dist && cross-env NODE_ENV=production webpack --progress --config ./webpack.config.js",
     "test": "echo 'No tests!'",
-    "lint": "eslint --fix src"
+    "lint": "eslint src"
   },
   "repository": {
     "type": "git",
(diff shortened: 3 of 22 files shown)


packages/budget/package.json
```diff
@@ -29,7 +29,7 @@
     "start:dev": "cross-env NODE_ENV=development node server",
     "start:prod": "cross-env NODE_ENV=production node server",
     "test": "cross-env BABEL_ENV=test mocha --opts ./mocha.opts 'src/**/*.test.js'",
-    "lint": "eslint --fix src"
+    "lint": "eslint src"
```
Dianna: I'm getting this error when running the command in this package:
```bash
dna budget: (lint-scripts)$ yarn run lint
yarn run v1.13.0
$ eslint src

Oops! Something went wrong! :(

ESLint: 5.16.0.
You are linting "src", but all of the files matching the glob pattern "src" are ignored.

If you don't want to lint these files, remove the pattern "src" from the list of arguments passed to ESLint.

If you do want to lint these files, try the following solutions:

* Check your .eslintignore file, or the eslintIgnore property in package.json, to ensure that the files are not configured to be ignored.
* Explicitly list the files from this glob that you'd like to lint on the command-line, rather than providing a glob as an argument.

error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

package.json
```diff
@@ -34,7 +34,7 @@
     "start": "node scripts/start",
     "deploy-storybook": "BABEL_ENV=esm storybook-to-ghpages",
     "in": "lerna run --scope",
-    "lint": "lerna run lint",
+    "lint": "lerna run lint --no-bail",
```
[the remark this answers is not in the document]
jaronheard: Yep, that would be the idea :)

package.json
```diff
@@ -48,11 +48,12 @@
     }
   },
   "lint-staged": {
-    "*.{js,jsx,json}": [
+    "*.{js,jsx}": [
       "prettier --write",
+      "eslint --fix",
```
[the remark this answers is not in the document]
jaronheard: Yes, that was the intent. I was hoping this would inspire a "clean as you go" approach. I was also hoping to have things in a pretty clean starting place. Hopefully it's not too annoying. Follow-up idea coming...

packages/budget/package.json
jaronheard: Ah, there's a .eslintignore file in that package that ignores ... `**/*.js`!

I wonder if there's a command eslint that will turn no-files into a pass. That way, we could enforce linting for the whole project at the CI level for all of the packages that have been linted, before all of the packages have been linted.

package.json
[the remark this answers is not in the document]
jaronheard: Actually, mis-read your question. Once we've got all packages reporting a pass on `lint`, we can run this as part of our CI tests. For now, it lets you see all remaining linting errors in all packages, without stopping the whole script the first time that it sees a linting error.

packages/budget/package.json
jaronheard: It looks like there's not a command to pass to eslint for this, but they're working on it (eslint/eslint#10587)

We could workaround, if we feel like going for it. A couple of reference examples:
- > We're currently shipping a file called eslint-no-cry.js in almost every seed, which have the only purpose of avoiding eslint cli to exit the build process when there is no .js file to be linted (eslint/eslint#10587).
- https://github.com/davidchambers/doctest/pull/111/files

package.json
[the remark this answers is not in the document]
Dianna: Cool. It is doing that. I'd recommend putting a card on the project board so it doesn't get lost. Plus it's always good to lower our bus factor (still, please don't die)

package.json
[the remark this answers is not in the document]
Dianna: Cool. Just checking :)

packages/budget/package.json
Dianna: Nifty. Do you want to do that or shall I add the commit?

jaronheard: All you :)

Dianna: Fixed! Since the local `.eslintignore` is used I _think_ we're good for a travis build, but just in case something goes wrong in building we can check out the `.travis.yml`.

Dianna: After you review my change I'll aprove

PR #535, opened 2019-05-14
Added README section for component library/project workflow

jaronheard:
Added a section to the README about working on the component library and a project package at the same time, following up on working with @kdv24 on #534

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ $ yarn test
 $ yarn test --watch
 ```
 
-## Working on the component library using storybook
+## Working on the component library using Storybook
 
 We are committed to a shared component library. This is achieved using the `component-library` package and React Storybook.
 Run Storybook with the following command or [view it here](https://hackoregon.github.io/civic/):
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -117,6 +117,15 @@ Run Storybook with the following command or [view it here](https://hackoregon.gi
 $ yarn storybook
 ```
 
+## Working on the component library and a project package simultaneuously
+
+In separate terminals, run the commands in both sections above. Project packages rely on the built version of the component library, so if you have updated the component library, and want to see your changes in the project package you are working on, you'll need to rebuild the component library. Once the component library build has finished, your project package will reload with the update components.
```
kdv24: It might be helpful to say which sections you mean when you say "both sections," just to be explicit.

jaronheard: Great suggestion! I've implemented in my most recent changes

jaronheard:
@kdv24 Thanks for the comments and review! I've updated per your suggestion. As per the `cd packages/2018` and `yarn start`, that's covered in the "Working on a single package other than the component-library".

A potential future area for improvement would be how a developer knows what package(s) to work on. Maybe this is something that could be documented in our issues or issue templates?

README.md
kdv24: 👍

Issue #463, opened 2019-04-10
Add more options to civicFormat utility


mendozaline:
Per Jaron in [#445](https://github.com/hackoregon/civic/pull/445#discussion_r273321487):
In the component library, add more options to the [civicFormat utility function](https://github.com/hackoregon/civic/blob/master/packages/component-library/src/utils/civicFormat.js).
- [x] Add option to abbreviate numbers over a thousand
- Ex. 2,000 => 2k
- [x] Optional: Add option to show percentages to one decimal point
- Ex. 0.3456 => 34.6%

mendozaline:
Addressed by #551

PR #551, opened 2019-05-21
added abbreviate for bil, mil, thousand

corydm92:
Addresses issue  #463 

Used functions described above to quickly test new abbreviation function.

Added decimalToPercent method, using the Number function and d3 format method.

### DIFF
--- a/packages/component-library/src/utils/civicFormat.js
+++ b/packages/component-library/src/utils/civicFormat.js
@@ -36,6 +36,41 @@ const numeric = d => {
   return formatted;
 };
 
+const scalesShort = [
+  [1000000000000, "t"],
+  [1000000000, "b"],
+  [1000000, "m"],
+  [1000, "k"]
+];
+
+const abbreviateLargeShort = number => {
+  let num;
+  let scale;
+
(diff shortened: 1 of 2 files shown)


packages/component-library/src/utils/civicFormat.js
```diff
@@ -36,6 +36,53 @@ const numeric = d => {
...
+    }
+  }
+
+  return `${num}${scale}`;
+};
+
+const numericShort = d => {
+  let formatted;
+
+  // We want to specifically format numbers greater than one million.
```
jaronheard:
```suggestion
  // We want to specifically format numbers greater than one thousand.
```

packages/component-library/src/utils/civicFormat.js
```diff
@@ -36,6 +36,53 @@ const numeric = d => {
...
+    formatted = abbreviateLargeShort(d);
+  } else {
+    formatted = format(",.0f")(d);
+  }
+
+  return formatted;
+};
+
+const decimalToPercent = num => {
+  let formated;
```
jaronheard: formated => formatted

PR #553, opened 2019-05-22
Design system/bar chart

bradfordjanice:


packages/component-library/stories/BarChart.story.js
```diff
@@ -1,48 +1,135 @@
...
+import civicFormat from "../src/utils/civicFormat";
 
-import { withKnobs, text, object } from "@storybook/addon-knobs";
+import {
+  withKnobs,
+  text,
+  object,
+  boolean,
+  number,
+  optionsKnob as options
```
jaronheard: I was curious about the naming of this knob -- It's unfortunate that Storybook has this knob with a different naming convention then all other knobs. Understandable, though.

bradfordjanice: Yes, they are using a different naming convention.  Here is the documentation for it:
https://github.com/storybooks/storybook/tree/master/addons/knobs

Scroll down to the "options" knob and it has an example of using this knob.

packages/component-library/stories/BarChart.story.js
```diff
@@ -1,48 +1,135 @@
...
+            { ye: 1996, population: 6000 },
+            { ye: 1997, population: 3000 },
+            { ye: 1998, population: 1000 }
+          ],
+          GROUP_IDS.DATA
+        );
+        const barWidth = number("Bar width", 37, {}, GROUP_IDS.CUSTOM);
+        const loading = boolean("Loading...", false, GROUP_IDS.CUSTOM);
+        const error = boolean("Error", false, GROUP_IDS.CUSTOM);
+        const getKeys = obj => {
```
jaronheard: A little bit cleaner:
```
const getKeyNames = obj => {
  const keyNames = {};
  Object.keys(obj).forEach(key => {
    keyNames[key] = key;
  });
  return keyNames;
};
```

jaronheard: Basically the same as before, just with cleaner naming

jaronheard: Also, this function definition should be moved out of .add to the main file scope

PR #564, opened 2019-05-23
Remove deprecated

kdv24:
Progress toward removing the ten components that are deprecated in the component-library.

Screen Shot 2019-05-23 at 4 44 47 PM

* Should these each have been their own PR?


jaronheard:
> Should these each have been their own PR?

@kdv24 I think that doing this all as one PR is fine!

kdv24:
### Components identified as Deprecated (√ once removed):
- [x] Chart.js
- [x] ChartData.js
- [ ] HeatMap.js - reverted with last commit b/c will be used going forward
- [x] HexOverlay.js
- [x] Media.js
- [x] Sankey.js
- [x] ShareCollection.js
- [ ] StoryFooter.js - used in StoryCard.js & CivicStoryCard
- [ ] StoryLink.js - used in StoryFooter.js
- [x] Tag.js

Okay, should be good to go now. @jaronheard will make issue cards for the StoryLink and StoryFooter components referenced above.

packages/component-library/stories/index.js
```diff
@@ -15,8 +15,6 @@ import dataTable from "./DataTable.story";
 import dropdownStory from "./DropdownMenu.story";
 import gradientScaleStory from "./GradientScale.story";
 import headerStory from "./Header.story";
-import heatMapStory from "./HeatMap.story";
```
jaronheard: Keep this line

packages/component-library/stories/index.js
```diff
@@ -135,8 +132,6 @@ storiesOf("Component Lib|Maps", module)
 baseMapStory();
 boundaryMapStory();
 civicSandboxMapStory();
-heatMapStory();
```
jaronheard: Keep this line

mendozaline:
Since heat maps are now created within the Base Map component, the original Heat Map component can be removed. Specifically this file here: 
https://github.com/hackoregon/civic/blob/master/packages/component-library/src/HeatMap/HeatMap.js

Also line 63 in this file can be removed as well: 
https://github.com/hackoregon/civic/blob/master/packages/component-library/src/index.js#L63

The Heat Map story and the lines in ` stories/index.js ` should be kept as noted by Jaron above.

kdv24:
@mendozaline because I hadn't realized someone else was working on the HeatMap and I neglected to pull again before today, I'm concerned that my more recent changes will push over the top of the changes you made to the Heat Map story itself.  I think it'd be worth copying what you did so it can be played back on top of mine.  I can do it myself, but I'd rather you got credit for the work you did!

packages/component-library/stories/index.js
kdv24: added back in

packages/component-library/stories/index.js
kdv24: yup. done. thx!

kdv24:
Thanks @mendozaline and @jaronheard!  I think this can be merged now.

PR #565, opened 2019-05-24
Design system/horizontal bar chart

bradfordjanice:
Updated the HorizontalBarChart story: organized stories, added notes, created knobs for each story.

### DIFF
--- a/packages/component-library/stories/HorizontalBarChart.story.js
+++ b/packages/component-library/stories/HorizontalBarChart.story.js
@@ -2,9 +2,28 @@ import React from "react";
 /* eslint-disable import/no-extraneous-dependencies */
 import { storiesOf } from "@storybook/react";
 import { checkA11y } from "@storybook/addon-a11y";
-import { withKnobs, text, object } from "@storybook/addon-knobs";
+import {
+  withKnobs,
+  text,
+  object,
+  boolean,
+  optionsKnob as options
+} from "@storybook/addon-knobs";
 import { HorizontalBarChart, civicFormat } from "../src";
 import notes from "./horizontalBarChart.notes.md";
+
+const GROUP_IDS = {
+  LABELS: "Labels",
+  DATA: "Data",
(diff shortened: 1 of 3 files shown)


packages/component-library/stories/HorizontalBarChart.story.js
```diff
@@ -2,122 +2,237 @@ import React from "react";
...
-  { population: 6000, label: "French Bulldog" },
-  { population: 3000, label: "Afghan Hound" },
-  { population: 1000, label: "Jack Russell Terrier" }
-];
+const GROUP_IDS = {
+  LABELS: "Labels",
+  DATA: "Data",
+  CUSTOM: "Custom"
+};
+const getKeyNames = obj => {
```
jaronheard: Since this is reused in BarChart, HorizontalBarChart, and will be used in other stories, extract to shared.js helper file and import rather than duplicating.

packages/component-library/stories/HorizontalBarChart.story.js
```diff
@@ -2,122 +2,237 @@ import React from "react";
...
-    })
-    .add("No title", () => {
-      const data = object("Data", sampleData);
-      const sortOrder = text("Data series", "sortOrder");
-      const dataValue = text("Data values", "population");
-      const dataLabel = text("Data series labels", "label");
+        return (
+          <HorizontalBarChart
+            data={data}
+            sortOrder={sortOrder}
```
jaronheard: Technically, sortOrder isn't required, the component will pick a sortOrder by default. We should move it to the custom story.

A follow up issue for this is that the default sort should be changed to sort from highest value to lowest value, rather than the current sorting method.

packages/component-library/stories/HorizontalBarChart.story.js
bradfordjanice: I will also extract getKeyNames from the BarChart story and import from shared.js.  (In a new branch.)

Issue #567, opened 2019-05-24
Design System - Update ScatterPlot Map Stories


mendozaline:
## What
**What needs to be done?**

**Standard**
- [x] Rename"Simple Usage" story to "Standard" story
- [x] Create knobs for each of these standard props and group as marker:
-`getFillColor`
-`getRadius`
-`opacity `
- [x] Create knobs for each of these standard props and group as data:
-`data`
-~~`getPosition`~~

**Custom**
- [x] Create "Custom" story 
- [x] Create knobs for each of these custom props and group as custom:
-`stroked`
-`getLineColor`
-`getLineWidth`
-`autoHighlight`
-`highlightColor`
-`onLayerClick`

**Example**
- [x] Rename "With tooltip" story to "Examples: With Tooltip" story

**Notes**
- [x] Add ScatterPlot Map Notes file

## Why
**Why does it need to be done? (Include links to related issues)**
Scatter Plot Map stories should follow the Civic design system pattern established in  [#508](https://github.com/hackoregon/civic/pull/508), [#553](https://github.com/hackoregon/civic/pull/553), and [#565](https://github.com/hackoregon/civic/pull/565). Please references these files:

**Button Component**
-[Button Stories File](https://github.com/hackoregon/civic/blob/master/packages/component-library/stories/Button.story.js)
-[Button Notes File](https://github.com/hackoregon/civic/blob/master/packages/component-library/stories/button.notes.md)

**Bar Chart Component**
-[Bar Chart stories](https://github.com/hackoregon/civic/blob/master/packages/component-library/stories/BarChart.story.js)
-[Bar Chart notes ](https://github.com/hackoregon/civic/blob/master/packages/component-library/stories/barchart.notes.md)

**Horizontal Bar Chart Component**
-[Horizontal Bar Chart stories](https://github.com/hackoregon/civic/blob/f7dcbc59a221df811d26d0133dd1b81929bb6761/packages/component-library/stories/HorizontalBarChart.story.js)
-[Horizontal Bar Chart notes](https://github.com/hackoregon/civic/blob/f7dcbc59a221df811d26d0133dd1b81929bb6761/packages/component-library/stories/horizontalBarChart.notes.md)

## When
**When does this need to be done?**
In the near future

paigewilliams:
[...]

PR #583, opened 2019-05-26
Design System - update scatterplot map story

paigewilliams:
Addresses #567 

Summary:

I updated the standard and with tooltip examples, and created the custom and data driven styling examples. I also added a notes file that describes the different props and explains each example in the story. 

I currently display the data array in the data tab in the knobs. Requested in the issue is to add `getPostion` knob. I did not add it because the user can see/edit the coordinates in the data array and the prop is explained in the notes file. 

Additionally, with the current configuration I receive prop-type warnings for `getLineWidth`, `getLineColor` and `getFillColor` because they are expecting functions, where I am returning arrays or numbers. According to the [deck.gl](https://github.com/uber/deck.gl/blob/master/docs/layers/scatterplot-layer.md) documentation you can supply either data types. We could either change the prop types to be `onOfType` or try to rewrite my code to supply functions. I tried to supply functions to these props but was having difficulty connecting them to the knobs. 

Looking forward to feedback 🚀

### DIFF
--- a/packages/component-library/stories/ScatterPlotMap.notes.md
+++ b/packages/component-library/stories/ScatterPlotMap.notes.md
@@ -17,9 +17,9 @@ The Custom story shows all the possible properties that can be passed to the Sca
 
 - `data` : expects an array of objects.
 - `getPosition` : expects a function that returns the coordinates of each object in data array.
-- `opacity` : expects a number.
-- `getFillColor` : expects a function or an array in `[r, g, b, [a]]` format. If an array is provided, the color is applied to all objects. If a function is provided, it is called on each object to set the fill color.
-- `getLineColor` : expects a function or an array in `[r, g, b, [a]]` format. If an array is provided, the color is applied to all objects. If a function is provided, it is called on each object to set the fill color.
+- `opacity` : expects a number between 0 and 1.
+- `getFillColor` : expects a function or an array in `[r, g, b, [a]]` format, where `a` is a number between 0 and 255. If an array is provided, the color is applied to all objects. If a function is provided, it is called on each object to set the fill color.
+- `getLineColor` : expects a function or an array in `[r, g, b, [a]]` format, where `a` is a number between 0 and 255. If an array is provided, the color is applied to all objects. If a function is provided, it is called on each object to set the fill color.
 - `getRadius` : expects a function or a number. If a number is provided it is used as the radius for all objects. If a function is provided, it is called on each object to set the radius.
 - `radiusScale` : expects a number.
 - `stroked` : expects a boolean value.
--- a/packages/component-library/stories/ScatterPlotMap.notes.md
+++ b/packages/component-library/stories/ScatterPlotMap.notes.md
@@ -1,5 +1,39 @@
-# HorizontalBarChart Component
+# ScatterPlot Map Component
 
 ## Standard
 
+The Standard story shows the standard usage of the ScatterPlot Map Component for the CIVIC platform. ScatterPlot Map includes the standard styling.
+
+These properties can be set in the standard story:
+
+- `getFillColor`
+- `radiusScale`
+- `opacity`
+- `data`
+
 ## Custom
+
+The Custom story shows all the possible properties that can be passed to the ScatterPlot Map Component for customization.
+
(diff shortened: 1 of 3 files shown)


packages/component-library/stories/ScatterPlotMap.notes.md
```diff
@@ -0,0 +1,39 @@
...
+- `opacity`
+- `data`
+
+## Custom
+
+The Custom story shows all the possible properties that can be passed to the ScatterPlot Map Component for customization.
+
+- `data` : expects an array of objects.
+- `getPosition` : expects a function that returns the coordinates of each object in data array.
+- `opacity` : expects a number.
```
mendozaline: It would be good to note here that the number to set the opacity should be between 0 and 1.

packages/component-library/stories/ScatterPlotMap.notes.md
```diff
@@ -0,0 +1,39 @@
...
+- `getFillColor` : expects a function or an array in `[r, g, b, [a]]` format. If an array is provided, the color is applied to all objects. If a function is provided, it is called on each object to set the fill color.
```
mendozaline: For all the color options, we should make clear that the numbers for each color channel (rgb) and for opacity (a) should be between 0 and 255.

mendozaline:
Great work, Paige! Your updates to the ScatterPlotMap story are a major improvement.

Don't worry about the `getPosition` knob. I think your notes make its purpose clear. Although in the notes, it should still be listed under Standard.  

Also I don't think the ScatterPlotMap component in the Standard story needs every prop passed to it. I think only the 5 props listed under Standard in the notes would suffice.

Your suggestion to update the ScatterPlotMap component's `propTypes` to to use `oneOfType` makes sense. Let's incorporates those changes here.

One question I have is whether we should be using the `color` knob. I used it in the Boundary Map story but I had to use that workaround to convert hex color codes into rgb color codes. While that works, I think this might confuse new contributors about what is actually being passed to the props that set color options for the map. Perhaps we should just be using the `object` knob instead and pass that directly to the props like `getFillColor`.

paigewilliams:
@mendozaline thanks for the feedback! I'll update the notes per your requests and refactor the Standard map to only receive the required props. I'll also update the ScatterPlotMap components `propTypes`. 

I do think that the color knob UI is nice and would be convenient for the developer to use. The notes do state the format of data required and the rgba values are visible when you click on the knob. I am happy to make the change if we want to have object knobs for rgba values across the board.

mendozaline:
I also like the convenience of the `color` knob. So how about if we keep it in the Custom and Example stories, but use the `object` knob in the Standard story.

bradfordjanice:
Another suggestion...use sample data that reflects the type of data CIVIC works with.  Perhaps look at one of the projects that uses a scatter plot and use a subset of that data.  It will reflect how the components are designed and used in CIVIC's platform.

paigewilliams:
@mendozaline @bradfordjanice Thank you for the suggestions! I will keep the color knob in the example and custom story, but change it to the object in the standard story.

bradfordjanice:
Hi Paige,

This is a great story!!  Editing these props and seeing the changes in the map will make designing/using this component much more transparent.  Also, I like that you added proptype information in the Notes.

I noticed a couple of things, not necessarily that require changes, but want to let you know.  Looking at the code for the Custom story, I think you mean for an action to show in the Ation tab of the addon panel when the map is clicked (per the onLayerClick prop), but I don't see anything when I click on the map.

Also, the Data Driven Styling example doesn't have a way for the user to adjust the value that drives the styling?  That would be very useful if it's possible.

paigewilliams:
@bradfordjanice Thank you for the thoughtful review! I believe that you are not seeing the action trigger in the standard story because I do not pass in `onLayerClick` as a prop in that story. I am intending to show the user the required props in that story, and `onLayerClick` is not required. 

Regarding adding a way for the user to adjust the data driven styling, I tried to add a knob for that and I couldn't find a way to do so in the time I alloted. I do agree that it would be a nice feature. Perhaps we could try it in a different ticket and implement it among more stories. 

@jaronheard I'll merge it right now!

Issue #447, opened 2019-04-02
Improvements to 2017 Emergency Response


jaronheard:
Capturing these follow-up comments on #442 for the 2017 Emergency Response project.

- [ ] Clicking an FMA doesn't highlight the polygon in the map, so it's hard to recall what is open.
- [ ] The FMA tooltip/details pane is off brand.
- [ ] The labels in the pie chart start to overlap.
- [ ] The points in small multiple charts exacerbate the known hover target size problem.
- [ ] The hover text for the small multiple charts are too small, since they scale with the chart.

PR #584, opened 2019-05-29
Design System - Map Overlay

mendozaline:
This addresses #571.

The MapOverlay component and stories were updated to conform to the new design system standard.

- Updated the MapOverlay component's props, propTypes, and defaultProps
- Renamed the MapOverlay stories to Standard, Custom, and Examples: With Tooltip
- Added a data knob to the Standard and Custom stories to show the first feature object from the GeoJSON passed to it
- Created an Examples: Choropleth Map story 
- Added a MapOverlay notes markdown file
- Updated components that use MapOverlay in transportation and housing packages
- I was also able to address an item from #447 by making the selected polygon on the FMA maps persistently highlighted

### DIFF
--- a/packages/component-library/src/MapOverlay/MapOverlay.js
+++ b/packages/component-library/src/MapOverlay/MapOverlay.js
@@ -1,32 +1,39 @@
 import React from "react";
-import PropTypes from "prop-types";
+import {
+  shape,
+  bool,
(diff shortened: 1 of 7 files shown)


mendozaline:
Thanks for the review, Jeannie!

Issue #558, opened 2019-05-23
Make yarn start open the browser


jaronheard:
### What

Make `yarn start` open the browser when the server starts in project packages, same as `yarn storybook`. This will be a change in the `dev-server` package, used by all other project packages.

### Why

Currently, running `yarn start` in a package doesn't open the browser, but `yarn storybook` does.  Changing this will make a consistent experience for development across different parts of CIVIC

### When

As soon as possible.

corydm92:
Assigned, currently opening in a new tab but not waiting for Webpack Bundle. Linting errors were bypassed using git commit --no-verify for now, will be looking at the linting issue in the future.

jaronheard:
#597 addressed

PR #597, opened 2019-06-06
Issue #558 open localhost

corydm92:
Doesn't wait for Webpack Bundle Analyzer to finalize before opening localhost.


packages/dev-server/index.js
```diff
@@ -76,5 +77,12 @@ module.exports = function() {
...
   const port = process.env.PORT || 3000;
-  app.listen(port, announceServer);
+
+  const setupServer = new Promise(function(resolve, reject) {
+    resolve(app.listen(port, announceServer));
+  });
+
+  setupServer.then(open(`http://localhost:3000`));
+
+  // Doesn't wait for Webpack Bundle Analyzer to finalize before opening localhost.
```
jaronheard: If this comment is accurate, and it’s finishing the Webpack bundle before starting, and just not finishing starting the Webpack Bundle Analyzer, I think that’s totally fine.

This is a good explanatory comment, so leave it in.

packages/dev-server/package.json
```diff
@@ -20,7 +20,11 @@
   "devDependencies": {
     "redux-logger": "^3.0.6",
     "webpack-dev-middleware": "^3.6.1",
-    "webpack-hot-middleware": "^2.22.0"
+    "webpack-hot-middleware": "^2.22.0",
+    "open": "^6.3.0"
```
jaronheard: After going through the process of upgrading dependencies, I’m more wary of adding new ones, but it looks like this a very well maintained package, and the same one used by Storybook.

corydm92:
Sounds great

Issue #498, opened 2019-04-30
Make Slider Component Standalone


jaronheard:
## What

The Slider component in the component library should not render a tooltip by default

### Why

In order to build #477, we need a simple slider.

### When

By May 6

PR #598, opened 2019-06-06
Slider component (Issue #498)

BJHunnicutt:
Issue #498: This simplifies and cleans up the Slider component to be more generic. Custom features that were only being used by 2017 housing were moved over to that folder as a CustomSlider.

(There are also a few unrelated lint fixes in here from running yarn lint --fix)


bradfordjanice:
(My comments are only for the Slider story only...)

The StatefulWrapper utility is nice!  A couple of suggestions...
Does it make sense to put StatefulWrapper in the src/utils folder with other helpers?  Also, there is a file src/utils.js that imports helper functions; it could be added to that if appropriate.

The Slider stories can be updated to follow the new pattern for the design system:

Include the data, knobs, and call to that component in .add, for each story:
... adding knobs outside of .add causes unexpected behavior between stories
... when looking at "Story" in the addon panel, the code for that story is highlighted.  The goal is to have all the data and props for that instance of the component highlighted.

Name the story that shows the most common use of the component "Standard".  The story that uses all of the knobs/props in Standard, plus any additional knobs/props is named "Custom".  "Example" stories show configurations for specific uses.

For consistency in Storybook, for the knob labels, use descriptive names in sentence case, instead of prop names.

Create Notes, slider.notes.md, that briefly describes the stories and props that can be set for each story.
(This may be modified in the near future with Storybook Docs.)

Is there anyway to add the value format as a knob in the Slider with tooltip story?  That would be very helpful to the developer... they would be able to see what formats are available and test them out.

Some Storybook stories that use the pattern for the design system are:
Bar Chart
Line Chart

packages/component-library/src/Slider/Slider.jsx
```diff
@@ -0,0 +1,63 @@
+import React from "react";
```
jaronheard: Change component extension from .jsx to .js for consistency?

packages/component-library/src/Slider/Slider.jsx
```diff
@@ -0,0 +1,63 @@
...
+  return (
+    <span className="civic-slider-container">
+      <SliderComponent
+        dots={dots}
+        min={min}
+        max={max}
+        step={step}
+        value={value}
+        onChange={onChange}
+        {...rest}
```
jaronheard: I haven’t used {...rest} but I guess this would allow us to pass in any of the additional props that RCSlider or a custom SliderComponent takes. Seems handy! One trade off is that if someone uses the props that aren’t documented it’d be harder to change the component implementation in the future. I think I like it, but are there other trade offs? 

@mendozaline have you thought about using {...rest} in our map components?

packages/component-library/src/Slider/Slider.jsx
```diff
@@ -0,0 +1,63 @@
...
+  </span>
+);
+
+Slider.propTypes = {
+  max: number.isRequired,
+  min: number.isRequired,
+  onChange: func.isRequired,
+  showTooltip: bool,
+  step: number,
+  dots: bool, // step markers
```
jaronheard: Maybe we could call this prop `showStepMarkers` since it was unclear enough that it warranted a comment?

packages/component-library/stories/helpers/StatefulWrapper.jsx
```diff
@@ -0,0 +1,32 @@
+import React, { Component } from "react";
```
jaronheard: Switch component extension from .jsx to .js for consistency

packages/component-library/stories/helpers/StatefulWrapper.jsx
```diff
@@ -0,0 +1,32 @@
...
+
+  // Getter for any value in state
+  get = stateKey => this.state[stateKey];
+
+  // Setter for any value in state
+  set = newStateObject => this.setState(newStateObject);
+
+  render() {
+    const { children } = this.props;
+    return <span>{children({ get: this.get, set: this.set })}</span>;
```
jaronheard: Can you use `<React.Fragment>` here instead of `<span>` to avoid adding an extra DOM element? (I’m not 100% sure this works but if you can it would be nice!)

packages/component-library/stories/helpers/StatefulWrapper.jsx
jaronheard: Also move to common location per @bradfordjanice’s comment

Issue #602, opened 2019-06-07
Convert 2017 components to use CivicStoryCard and remove StoryCard


jaronheard:
### What

2017 projects should be converted to use CivicStoryCard rather than StoryCard.

- [ ] Housing
- [x] Transportation
- [x] Homelessness
- [x] Budget
- [x] Emergency Response

2017 project cards should have a blank watermark and not show a footer. (`<CivicStoryCard watermark={<div/>} footer={false}`)

**Why no watermark?** The 2017 cards weren't designed with the watermark in mind and aren't built to the same standards as the 2018 and later cards.

**Why no footer?** The 2017 package don't have the same card-registry as the 2018 package that allows cards to have their own page.

### Why

So we can remove StoryCard, StoryCardFooter, and StoryCardLink from the component library per #545.

### When

Once we have a pattern for CivicStoryCard that can accommodate the cards from the 2017 projects.

One way to accommodate these cards would be to remove the watermark and the footer.

jaronheard:
#2017 addressed

Issue #545, opened 2019-05-20
Remove unused deprecated components


jaronheard:
### What

Remove unused components marked as "deprecated" from component-library.

<img width="286" alt="image" src="https://user-images.githubusercontent.com/7065695/58055556-ec900500-7b12-11e9-88a6-0a4f1a58a6ab.png">

Please ensure the following for each deprecated component:
- [ ] The component in question is not imported from the component-library by any other packages
- [ ] Remove all files (source, test, stories) and references to component from the component-library

If a deprecated component is used in a project package, please file an issue to upgrade that component.

### Why

Removing unused components will eliminate developer confusion and trim the size of our bundles

### When

Not time sensitive

### Components identified as Deprecated (√ once removed):
- [x] Chart.js
- [x] ChartData.js
- [x] HeatMap.js
- [x] HexOverlay.js
- [x] Media.js
- [x] Sankey.js
- [x] ShareCollection.js
- [x] StoryFooter.js - used in StoryCard.js & CivicStoryCard
- [x] StoryLink.js - used in StoryFooter.js
- [x] Tag.js

kdv24:
@jaronheard maybe I'm missing something, but when I search for "deprecated" in civic I'm not getting any hits...

jaronheard:
@kdv24 maybe case sensitivity -- https://github.com/hackoregon/civic/search?q=deprecated&unscoped_q=deprecated

kdv24:
yep- @jaronheard that did it.  Thanks!

kdv24:
@jaronheard I removed the HeatMap component because it wasn't being referenced by any other components, but as I understand it now, changes were made so it _can_ be used going forward. It seems to me that I should amend my change to not remove it, and leave @mendozaline's changes.  Is there anything different that should be done with the HeatMap component?

kdv24:
Okay, should be good to go now. @jaronheard will make issue cards for the StoryLink and StoryFooter components referenced above.

jaronheard:
@kdv24 I reviewed your changes in #564, just a few lines to change. 

Per your comment above, I made issue #602 for the StoryLink and StoryFooter ... and then went ahead and did them myself in #604. If you're able to review #604 that would be ✨awesome✨!

kdv24:
Looks good. And, thanks for catching the ShareCollection bit I missed too!

PR #604, opened 2019-06-07
Convert 2017 components to use CivicStoryCard and remove StoryCard

jaronheard:
This addresses #602, and should close out #545.

For reference, here's what the 2017 projects look like now. See #602 for details on why no watermarks and footers.

Screen Shot 2019-06-07 at 14 51 26
Screen Shot 2019-06-07 at 14 50 35
Screen Shot 2019-06-07 at 15 33 45
Screen Shot 2019-06-07 at 15 33 56
Screen Shot 2019-06-07 at 15 34 04


packages/budget/src/components/CardCollection/index.js
```diff
@@ -19,11 +20,11 @@ const CardCollection = () => (
           decision-making!
         </p>
       </div>
-      <div className="ten-year__wrapper">
```
BJHunnicutt: There are some styles in CardCollection.css for `.ten-year__wrapper`, are they needed? If not, want to remove them?

packages/budget/src/components/Budget101/index.js
```diff
@@ -1,15 +1,14 @@
...
 import DataTable from "./DataTable";
-import { BUDGET_101 } from "../constants";
 
 import "./Budget101.css";
 import "./stub.css";
 
 function Budget101() {
   return (
-    <StoryCard title="Budget 101" collectionId="budget" cardId={BUDGET_101}>
+    <CivicStoryCard watermark={<div />} footer={false} title="Budget 101">
```
BJHunnicutt: This totally works, but I could see this `<div />` being confusing down the road. Would it make more sense to instead add an optional `hideWatermark` prop to CivicStoryCard?

BJHunnicutt:
Ran it locally and everything seems to work for me! Just left a couple optional comments

packages/budget/src/components/CardCollection/index.js
jaronheard: yep!

packages/budget/src/components/Budget101/index.js
jaronheard: I thought about this, and thought that leaving the `watermark` prop was more flexible without adding additional props to CivicStoryCard

PR #614, opened 2019-06-16
StackedAreaChart Stories

bradfordjanice:
This update to the Scatterplot component follows the Civic design system pattern established in #508, #553, and #565.

Working with the existing stories the following stories were created:

Standard - the common, default use of the StackedAreaChart component in the CIVIC platform
Custom - add knobs for all props of the StackedAreaChart component
Examples - examples of the component that show non-standard, but useful, configurations of the component:
The simple story shows only the most basic props.

A notes file was added; it describes each story and list the props/knobs that can be changed.

### DIFF
--- a/packages/component-library/stories/StackedAreaChart.story.js
+++ b/packages/component-library/stories/StackedAreaChart.story.js
@@ -6,7 +6,6 @@ import { storiesOf } from "@storybook/react";
 import {
   object,
   text,
-  array,
   withKnobs,
   optionsKnob as options
 } from "@storybook/addon-knobs";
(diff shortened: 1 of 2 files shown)


packages/component-library/stories/StackedAreaChart.story.js
```diff
@@ -104,87 +69,227 @@ const customLegend = legendData => {
...
           title={title}
           xLabel={xLabel}
           yLabel={yLabel}
-          legendComponent={customLegend}
+          xNumberFormatter={x => civicFormat[optionSelectX](x)}
+          yNumberFormatter={y => civicFormat[optionSelectY](y)}
         />
       );
-    });
+    })
```
jaronheard: Add notes

packages/component-library/stories/StackedAreaChart.story.js
```diff
@@ -104,87 +69,227 @@ const customLegend = legendData => {
...
+          { year: 2012, students: 55, group: "MULTI" },
+          { year: 2013, students: 56, group: "MULTI" },
+          { year: 2014, students: 59, group: "MULTI" }
+        ];
+        const enrollmentDataSeries = "group";
+        const enrollmentDataSeriesLabels = [
+          { category: "BAA", label: "Black/African American" },
+          { category: "HIS", label: "Hispanic/Latino" },
+          { category: "MULTI", label: "Multi-Ethnic" }
+        ];
```
jaronheard: If we're going to use this sample data, I'd like to include the source of the demographic group labeling in the subtitle.

For example, demographic categories designated by PPS, or census

bradfordjanice: Great suggestion.  I used a subset of the data in the Students from Historically Underrepresented Groups chart here:  http://civicplatform.org/cities/portland/neighborhood

But, it doesn't give the source of the demographic group labeling here.  Where would I find it?

jaronheard: Suggestion: just add "using selected Census categories"

PR #618, opened 2019-06-19
Check for invalid data props and display code instead of passing error to VictoryCharts

jaronheard:
This fixes a bug that was undocumented. 

If you pass a `dataKey`, `dataValue`, or `dataSeries` that doesn't correspond to properties of `data` to one of our chart components -- most commonly by adjusting Storybook knobs in the Data tab, then the chart would break, and wouldn't display properly even if you passed it valid properties after that. This made the Data knobs section of Storybook much less useful. 

For an example -- in the [BarChart story](https://hackoregon.github.io/civic/?path=/story/component-lib-charts-bar-chart--standard), under the Data knobs section, change Data key from `ye` to `year` and then back to `ye`. The BarChart no longer responds to valid data.

This fixes that bug, and displays a useful message showing the code when the data props.

Example:
image

One improvement I haven't made would be to refactor `allDataPropertiesExist` into a common helper function.


packages/component-library/src/BarChart/BarChart.js
```diff
@@ -31,8 +32,13 @@ const BarChart = ({
   error
 }) => {
   const chartDomain = domain || getDefaultDomain(data, dataKey, dataValue);
+  const allDataPropertiesExist =
```
jaronheard: Extract this logic to utils — 

`checkDataKeys` function that takes `( data, keysArray )` and returns an object `{ allKeysValid: bool, keyName1: { isValid, validKeys, invalidKeys } keyName2: { isValid, validKeys, invalidKeys } }`

packages/component-library/src/utils/DataChecker.js
```diff
@@ -0,0 +1,74 @@
...
+        {JSON.stringify(data, undefined, 2)}
+      </pre>
+    </div>
+  );
+  return component;
+};
+
+DataChecker.propTypes = {
+  data: PropTypes.arrayOf(PropTypes.shape({})),
+  dataAccessors: PropTypes.arrayOf(PropTypes.string),
```
mendozaline: I think this could be updated to `PropTypes.shape({})` since it's not an array but an object as it's currently implemented.

packages/component-library/src/utils/DataChecker.js
```diff
@@ -0,0 +1,74 @@
...
+        const validCount = results[values[index]].valid;
+        const totalCount = results[values[index]].total;
+        const keyValue = values[index];
+        const invalidStyle = !isValid
+          ? css`
+              color: red;
+            `
+          : css``;
+        return (
+          <pre className={invalidStyle}>
```
mendozaline: A `key` attribute should be added here to the `pre` element.

packages/component-library/src/utils/checkData.test.js
```diff
@@ -0,0 +1,103 @@
...
+      )
+    ).to.eql({
+      allKeysValid: false,
+      error: false,
+      keyType: { valid: 2, total: 2 },
+      lockType: { valid: 2, total: 2 },
+      category: { valid: 0, total: 2 }
+    });
+  });
+});
```
mendozaline: Nice tests!

packages/component-library/src/utils/DataChecker.js
jaronheard: Good catch!

packages/component-library/src/utils/DataChecker.js
jaronheard: Good catch

packages/component-library/src/utils/checkData.test.js
jaronheard: Thanks! I feel like they actually helped me do it faster, which is always a nice feeling.

Issue #616, opened 2019-06-18
Civic Sandbox Map 2.0


mendozaline:
### What
**What needs to be done?**
- [x] Update the Civic Sandbox Map component and props as needed
- [x] Update the Civic Sandbox Map story to include an example of each of these map types:
- Path Map
- ScatterPlot Map
- Screen Grid Map
- Icon Map
- Small Polygon Map
- Choropleth Map
- [x] Create a "Standard" tab for each map type
- [x] Create a "Custom" tab for each map type
- [x] Create a "Data" tab for each map type
- [x] Create knobs for each standard prop
- [x] Create knobs for each custom prop
- [x] Add a Civic Sandbox Map notes file

### Why
**Why does it need to be done? (Include links to related issues)**
The Civic Sandbox Map has to work with the new Sandbox API.

### When
**When does this need to be done?**
Soon

mendozaline:
Addressed by #619

PR #619, opened 2019-06-20
Design system/civic sandbox map

mendozaline:
This addresses #616.

It implements a new CivicSandboxMap component that will work with the new Sandbox API. Instead of directly updating the CivicSandboxMap component, I created a new component because if I changed the CivicSandboxMap it would break the current Sandbox.

The MultiLayerMap component will eventually replace the CivicSandboxMap once the new Sandbox API is ready.

This PR also adds a new MultiLayerMap story that conform to the new design system pattern. But it implements Standard and Custom stories as different tabs within each map story and not separate stories. Since the MultiLayerMap story will help the Sandbox team with layer creation, I thought this would be simpler for that purpose. I also added a notes file that describes the various props and map types. I opted to create the notes file as a template string so I could keep the code DRY.


mendozaline:
> MultiLayerMap looks great -- it looks flexible enough to implement outside of Sandbox. Was that one of your goals?

Yes, when someone wants to create a map with different layers on the same Base Map, this is how they can do that. The `MultiLayerMap` shouldn't be exclusive to the CIVIC Sandbox.


> The one thing I've noticed is that we've got some duplication going on with existing components -- I can see the MultiPathMap and the PathOverlay are both implementations of the deck.gl PathLayer. The "Multi" components are purpose-built to serve the MultiLayerMap - this makes sense, but this ends up with more components to support.

I'll start thinking about possible solutions that will allow us to use the map layers in a way that can allow for implementation in both stand alone maps and the CIVIC Sandbox. That way we don't have to have this much duplication.

PR #620, opened 2019-06-20
Design system - PieChart stories

bradfordjanice:
This update to the PieChart component follows the Civic design system pattern established in #508, #553, and #565.

Working with the existing stories the following stories were created:

Standard - the common, default use of the PieChart component in the CIVIC platform
Custom - add knobs for all props of the PieChart component
Example: Simple - the simple story shows only the most basic props.

A notes file was added; it describes each story and list the props/knobs that can be changed.

### DIFF
--- a/packages/component-library/stories/LineChart.story.js
+++ b/packages/component-library/stories/LineChart.story.js
@@ -277,7 +277,7 @@ export default () =>
       "Example: Many data points",
       () => {
         const scale = 0.25;
-        const value = number("Number of data points", 100, GROUP_IDS.DATA);
+        const value = number("Number of data points", 100, {}, GROUP_IDS.DATA);
         return (
           <LineChart
             data={Array(value)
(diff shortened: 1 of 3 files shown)


bradfordjanice:
I made the requested changes.  Adding "Data label" and "Data value" labels to the Standard story makes it consistent with the other chart component stories!  

The reason I had not added it was because it wasn't used in the pie charts I found in the projects or in the existing PieChart story.  The "Example: Simple" story shows the component using the default prop values for "Data label" and "Data value"; that seems the appropriate place.

After making these changes, the "Start angle" and "End angle" knobs don't work.  It was working, and I can't find the problem.  Looking at the component, it sets the value of startAngle and endAngle based on whether the prop halfDoughnut is true or false, not on the value of the knob.  I'm not sure why it was working or why it isn't working after these changes... if you could take a look I'd appreciate it.

If I'm right that these knobs aren't currently changing the value of startAngle and endAngle, we could either delete them or update the component.  By the way, none of the pie charts in the project have used startAngle and endAngle; perhaps it isn't useful.

jaronheard:
Awesome! I would remove the startAngle and endAngle knobs, and also remove the props in the component, since they aren't being used.

Jaron Heard
Creative Director
Civic Software Foundation

+1-971-998-7180

bradfordjanice:
In the last commit the "Start angle" and "End angle" knobs were removed from the Custom story.  The Notes were updated with this change.  The component was not changed.

Issue #474, opened 2019-04-14
Add Stacked Horizontal Bar Chart Option 


mendozaline:
The Civic Sandbox dashboard needs a stacked horizontal bar chart component. This will build off the Horizontal Bar Chart already in the component library.

It should look similar to this example, but horizontal: https://formidable.com/open-source/victory/gallery/100-column-chart/

- [x] Add stacked option to Horizontal Bar Chart
- [x] Add stacked example to Horizontal Bar Chart story

jaronheard:
#622 addressed

PR #622, opened 2019-06-21
Stacked horizontal bar chart

paigewilliams:
Addresses #474

**Summary**
I have added the `stacked` prop to the `HorizontalBarChart` component. I have also added a function to transform the data into 100% values, based on the `dataValue` and `dataLabel` provided. Additionally, I have added a function to dynamically generate the color scheme based on the number of arrays in the data provided, and assign them a color in the `CivicVictoryTheme.civic.group.colorScale` based on their index. I also added a `VictoryAxis` if the user passes in the `stacked` prop where the `tickFormat` is based on the `dataLabel`. 

There is an additional story to show what the stacked horizontal bar chart looks like and notes explaining how it works.


packages/component-library/stories/HorizontalBarChart.story.js
```diff
@@ -225,4 +225,50 @@ export default () =>
...
+            data={data}
+            dataValueFormatter={x => civicFormat[optionSelectX](x)}
+            dataValue={dataValue}
+            dataLabel={dataLabel}
+            xLabel={xLabel}
+            yLabel={yLabel}
+            title={title}
+            subtitle={subtitle}
+            minimalist={minimalist}
+            stacked
```
BJHunnicutt: I tried making 'stacked' a Knob and the chart wouldn't render anything anymore (I assume because the data format also needs to change). Generally, I feel like a single prop shouldn't be dependent on other changes to work. Would it make sense to either make a separate StackedBarChart component, or to make sure a non-stacked array of arrays renders as a grouped bar chart (https://formidable.com/open-source/victory/docs/victory-group/)?

packages/component-library/src/HorizontalBarChart/HorizontalBarChart.js
```diff
@@ -50,6 +52,32 @@ const HorizontalBarChart = ({
 
   const minValue = Math.min(0, ...data.map(d => d[dataValue]));
 
+  const categoricalColors = dataLength => {
```
BJHunnicutt: This seems like it could be a useful helper for other stories, could it go in a utils folder?

packages/component-library/stories/HorizontalBarChart.story.js
paigewilliams: I lean towards creating a new component because it requires an array of arrays of objects, rather than an array of objects. I think it could be misleading if we rendered a different type of chart based on the data structure, even if the user passes the `stacked` prop. I am open to other suggestions though!

packages/component-library/src/HorizontalBarChart/HorizontalBarChart.js
paigewilliams: I totally agree! Will do.

packages/component-library/stories/HorizontalBarChart.story.js
bradfordjanice: I suggest looking at the Stacked Area Chart and comparing it to the Bar Chart / Stacked Bar Chart.  If users, such as Data Scientists, think about how these two types of charts present data in a similar way (and from my research there is a lot of similarity), it would be good to design the stories, and the data structure for component, the same way.  It will improve usability.

For example, the Stacked Area Chart doesn't have a knob to say "stacked" or not, it's determined by the data file.  Compare the knobs in the Data tabs for the "Standard" and "Example: Simple" stories.  Should this change to be similar to the Bar Chart / Stacked Bar Chart, or should the Bar Chart / Stacked Bar Chart be similar to the Stacked Area Chart?

paigewilliams: @bradfordjanice Thanks for the advice, I will be sure to checkout that component and use my best judgement on how to proceed.

paigewilliams:
Changes since 7/20:

1. Create `groupBy` function that accepts an array of objects and groups them into arrays based on a common key/value. This function has tests as well.

2. Use the `groupBy` function in the `HorizontalBarChart` component and update props so that this is functioning. 

3. Update the story for the `HorizontalBarChart` to use this new functionality.

packages/component-library/stories/HorizontalBarChart.story.js
```diff
@@ -225,4 +225,68 @@ export default () =>
...
+          { name: "Poodle", x: "2018", y: 2 },
+          { name: "Poodle", x: "2019", y: 3 },
+          { name: "Bulldog", x: "2017", y: 2 },
+          { name: "Bulldog", x: "2018", y: 5 },
+          { name: "Bulldog", x: "2019", y: 4 },
+          { name: "Terrier", x: "2017", y: 1 },
+          { name: "Terrier", x: "2018", y: 2 },
+          { name: "Terrier", x: "2019", y: 3 }
+        ];
+        const dataValue = text("Data value", "x", GROUP_IDS.DATA);
```
jaronheard: _(Leaving this note on the story, but this requires changes to the component)_

Since this is the attribute used for the label displayed on the chart, I would have this be the `dataLabel` prop.

packages/component-library/stories/HorizontalBarChart.story.js
```diff
@@ -225,4 +225,68 @@ export default () =>
...
+        const dataLabel = text("Data label", "y", GROUP_IDS.DATA);
+        const groupByValue = text(
```
jaronheard: _(Leaving this note on the story, but this requires changes to the component)_

Since this is the attribute used for the category displayed on the chart, I would have this be the `dataSeries` prop, similar to our other components.

packages/component-library/stories/HorizontalBarChart.story.js
```diff
@@ -225,4 +225,68 @@ export default () =>
...
+          { name: "Poodle", x: "2019", y: 3 },
+          { name: "Bulldog", x: "2017", y: 2 },
+          { name: "Bulldog", x: "2018", y: 5 },
+          { name: "Bulldog", x: "2019", y: 4 },
+          { name: "Terrier", x: "2017", y: 1 },
+          { name: "Terrier", x: "2018", y: 2 },
+          { name: "Terrier", x: "2019", y: 3 }
+        ];
+        const dataValue = text("Data value", "x", GROUP_IDS.DATA);
+        const dataLabel = text("Data label", "y", GROUP_IDS.DATA);
```
jaronheard: _(Leaving this note on the story, but this requires changes to the component)_

Since this is the attribute used for the value displayed on the chart, I would have this be the `dataValue` prop.

packages/component-library/src/HorizontalBarChart/HorizontalBarChart.js
```diff
@@ -115,53 +138,85 @@ const HorizontalBarChart = ({
...
+              x="sortOrder"
+              y="dataValue"
+              events={chartEvents}
+            />
+          )}
+          {stacked && (
+            <VictoryStack colorScale={categoricalColors(groupedData.length)}>
+              {groupedData.map((arr, i) => {
+                return (
+                  <VictoryBar
```
jaronheard: Needs a `labelComponent` to show a tooltip, similar to non-stacked version.

packages/component-library/src/HorizontalBarChart/HorizontalBarChart.js
```diff
@@ -115,53 +138,85 @@ const HorizontalBarChart = ({
...
+              style={{
+                data: { fill: "none" }
+              }}
+              title="Horizontal Bar Chart"
+              x="sortOrder"
+              y="dataValue"
+              events={chartEvents}
+            />
+          )}
+          {stacked && (
```
jaronheard: We should use a `SimpleLegend` here similar to the `StackedAreaChart`

paigewilliams:
@jaronheard  I added the tooltip and legend per your requests! Let me know if there is anything else I should change!

jaronheard:
@paigewilliams I was taking a look, but it looks like maybe the SimpleLegend didn't get added in https://github.com/hackoregon/civic/pull/622/commits/ac21414c24b548785647a9b8a66bb70780627533 -- can you check it out? 

I see the tooltip working, and that's awesome! Going to hold off on the rest of the review in case there is anything else that didn't make it in.

paigewilliams:
Should be good to go!

packages/component-library/stories/HorizontalBarChart.story.js
```diff
@@ -225,4 +225,68 @@ export default () =>
         );
       },
       { notes }
+    )
+    .add(
+      "Example: Stacked bar chart",
+      () => {
+        const sampleStackedData = [
```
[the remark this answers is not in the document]
jaronheard: I'm still seeing `x` and `y` here -- and I believe that there's an issue with passing in data that has property names other than `x` or `y`, as when I tried to change it to `year` and `count` in the story I ran into NaN errors.

This would make sense, because `x` and `y` have special meanings in VictoryCharts. That's why there's uglier data mapping going on for the non-stacked components. But maybe it's possible to do with the `key` prop?

packages/component-library/stories/horizontalBarChart.notes.md
```diff
@@ -28,3 +28,13 @@ The Custom story shows all possible properties that can be passed to the BarChar
 ## Example: Minimalist
 
 ## Example: With Negative Values
+
+## Example: Stacked Bar Chart
+
+The Stacked Bar Chart shows all the possible properties for the stacked option of the Horizontal Bar Chart.
+
+The Stacked Bar Chart requires the `stacked` and `groupByValue` props. The `groupByValue` is the value that you would like to data to be grouped by. This is the value that is represented by the color of the bar, not the value of the x or y axis. This value is needed because the chart requires an array of arrays of objects, and the standard data structure for CIVIC is an array of objects.
```
jaronheard:
```suggestion
The Stacked Bar Chart requires the `stacked` and `dataSeries` props. The `dataSeries` is the value that you would like to data to be grouped by. This is the value that is represented by the color of the bar, not the value of the x or y axis.
```

jaronheard: Removed the last sentence since it's an implementation detail not relevant to the component user

jaronheard:
@paigewilliams Legend looking great, and overall this is looking great! 

This ended up in a comment, not in my review, but this is the main remaining issue: https://github.com/hackoregon/civic/pull/622#discussion_r305699936

PR #625, opened 2019-06-26
Feature orb component

lucastswick:
Adds `OrbManager` to control movement of `OrbComponents`


packages/project-disaster-trail/src/components/Game/OrbManager.js
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  const [models, setModels] = useState(tempModels);
+
+  useEffect(() => {
+    let animationFrameId;
+
+    const animate = () => {
+      tempModels = [];
+      for (let i = 0, model; i < count; i++) {
```
Dianna: Thinking of future devs (which may well be us in months or devs who've never done animation), it would be nice to have a note above the sections within this for loop, even something as simple as `//move onto screen` (for the if statements below) or `// orb bounce` for the model.y

PR #632, opened 2019-06-28
Feature active state

Dianna:
Adds the active state for the orbs. Example above the orb manager.
press-orb

### DIFF
--- a/packages/project-disaster-trail/src/components/Game/Orb.js
+++ b/packages/project-disaster-trail/src/components/Game/Orb.js
@@ -67,7 +67,7 @@ class Orb extends Component {
       window.cancelAnimationFrame(animationFrame);
       clearTimeout(pressTimeout);
       /* We keep setState() in a conditional in case of a mouseLeave when the orb had not been pressed to avoid unnecessary setState() calls */
-      this.setState(() => defaultState);
+      this.setState(defaultState);
     }
   };
 
@@ -79,12 +79,12 @@ class Orb extends Component {
     });
     const pressTimeout = setTimeout(this.handleOrbRelease, durationRequired);
 
-    this.setState(() => ({
+    this.setState({
       animationState: "pressing",
       pressedStart: new Date(),
       animationFrame: newAnimationFrame,
       pressTimeout
-    }));
+    });
   };
 
   doFullDurationAnimation = () => {
--- a/packages/project-disaster-trail/src/components/Game/Orb.js
+++ b/packages/project-disaster-trail/src/components/Game/Orb.js
@@ -14,8 +14,6 @@ const circleDefaultStyle = css`
   left: 10px;
   z-index: 10;
   background-color: mediumSeaGreen;
-  -webkit-transition: background-color 1000ms linear;
-  -ms-transition: background-color 1000ms linear;
   transition: background-color 1000ms linear;
   cursor: pointer;
   opacity: 0.8;
@@ -26,8 +24,6 @@ const circleDefaultStyle = css`
 
   &.circle-press-style {
     background-color: mediumAquamarine;
-    -webkit-transition: background-color 1000ms linear;
-    -ms-transition: background-color 1000ms linear;
     transition: background-color 1000ms linear;
   }
 `;
--- a/packages/project-disaster-trail/src/components/Game/Orb.js
+++ b/packages/project-disaster-trail/src/components/Game/Orb.js
@@ -5,9 +5,6 @@ import RadialGauge from "./RadialGauge";
 
 const orbContainerStyle = css`
   position: relative;
-  height: 80px;
-  width: 80px;
-  border-radius: 80px;
 `;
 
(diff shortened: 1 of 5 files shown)


packages/project-disaster-trail/src/components/Game/Orb.js
```diff
@@ -1,37 +1,46 @@
...
-`;
 
-const pressStyle = css`
-  background-color: #000;
-  -webkit-transition: background-color 1000ms linear;
-  -ms-transition: background-color 1000ms linear;
-  transition: background-color 1000ms linear;
+  &.circle-press-style {
+    background-color: mediumAquamarine;
+    -webkit-transition: background-color 1000ms linear;
```
lucastswick: I'm not sure but webpack should be autoprefixing these for us. Could you try removing `-webkit-transition` and `-ms-transition` rom `pressStyle` and confirming the prefixes still exist in the browser? If not, we'll want to update webpack so we don't have to write these out every time we need them. :)

packages/project-disaster-trail/src/components/Game/Orb.js
```diff
@@ -1,37 +1,46 @@
...
+    -ms-transition: background-color 1000ms linear;
+    transition: background-color 1000ms linear;
+  }
 `;
 
 const defaultState = {
   durationRequired: 1000,
+  orbSize: 60,
```
lucastswick: Woohoo!

packages/project-disaster-trail/src/components/Game/RadialGauge.js
```diff
@@ -0,0 +1,78 @@
...
+      if (animateGauge) {
+        this.setPercent(100);
+      } else {
+        this.setPercent(0);
+      }
+    }
+  }
+
+  setPercent = percent => {
+    this.setState(() => ({
```
lucastswick: I think this could be shortened to just `this.setState({percent})`

packages/project-disaster-trail/src/components/Game/RadialGauge.js
```diff
@@ -0,0 +1,78 @@
...
+
+    return (
+      <CircularProgressbar
+        value={percent}
+        strokeWidth={animateGauge ? 20 : 4}
+        css={css`
+          ${gaugeDefaultStyle}, ${gaugeSizeStyle}
+        `}
+        className={animateGauge ? "gauge-animate-style" : ""}
+        styles={buildStyles({
```
lucastswick: if this isn't changing per render, let's set a const outside the class to calculate it once, instead of on every render.

packages/project-disaster-trail/src/components/Game/RadialGauge.js
Dianna: I was never quite clear on how much [asynchronicity](https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous) would be an issue for this which is why I normally use the functional approach. My best guess would be it only matters if you're relying on the previous state / props which I'm not here (or the other places I use setState in this PR). Anyway, do you have any knowledge on that?

packages/project-disaster-trail/src/components/Game/Orb.js
Dianna: Ha! Just making it arbitrarily a bit bigger for ease of development ;)

packages/project-disaster-trail/src/components/Game/Orb.js
Dianna: removed

packages/project-disaster-trail/src/components/Game/RadialGauge.js
Dianna: Good catch. Changed it

packages/project-disaster-trail/src/components/Game/RadialGauge.js
lucastswick: All that `async` in terms of `setState` means is that the state is not set immediately. It's set at the end of the current frame. So if you:
```
// in your constructor
this.state = {progress: 0};

// later, in a function
this.setState({progress: 1});
console.log(this.state.progress); // returns 0, not 1.
```

Sidenote: 
You can think of setState as more of a queue. Like, hey I want 'progress' to change - and react says Yup Cool, give me a second to see if anything else is changing. Then it gathers _all_ the setStates and sets them once at the very end of the frame, instead of updating the state a bunch of times during the frame. Remember, every change in state triggers a render, so this is really just to minimize the number of renders that occur to a single one.

At any rate both the setPercent method here as well as the function inside the setState are called immediately, so there's no need to wrap `setState` or pass it a function to derive the `progress` - you should just be able to use `this.setState({progress});` instead of `setPercent`.   

Let me know if you have any other questions about state.

Dianna: I've removed the other functional calls for `setState()` :)

Issue #574, opened 2019-05-25
BarChart component:  Change the default sort order to be descending by dataValue


bradfordjanice:
# What
What needs to be done?

In the BarChart component, change the default sort order to be descending by dataValue.  See the sortOrder prop of the VictoryBar component.

Update the Custom BarChart story, with a knob that lets you select from descending or ascending options.

# Why
Why does it need to be done?

The BarChart component should provide an intuitive, default sorting option.  

# When
When does this need to be done?

In the near future

PR #640, opened 2019-07-08
Issue #574 sort order

corydm92:
The issue for creating ascending and descending sort options for the horizontal bar chart.


jaronheard:
Hey @corydm92 -- the code in here looks good, but it looks like some additional commits were included in this PR. 

It looks to me like the only commits that should be in this PR are these:
image

To remedy, can you do the following?
- Make sure your master branch is up to date
- Check out a new branch from master
- `git cherry-pick e20d1d4 698df21` to get just those two commits in your new branch. If there are other commits that should be included, include their commit ids as an argument to cherry-pick as well)
- Close this pull request and create a new pull request from your new branch

PR #643, opened 2019-07-10
Disaster/redux

lucastswick:
- Add redux support
- Add route support
- Prefer `<Link />` over `<a />`


packages/project-disaster-trail/src/components/LandingPage/index.js
```diff
@@ -4,10 +4,9 @@ import { Link } from "react-router";
 import styled from "styled-components";
```
jaronheard: Can we import `styled` from emotion instead of styled components? Trying to remove styled components as a dependency. Not urgent.

jaronheard: Also could use emotion css prop, instead if preferred.

lucastswick: You betcha.

Issue #631, opened 2019-06-28
Design System - Base Map


mendozaline:
**What needs to be done?**
**Standard**
- [x] Rename"Simple usage" story to "Standard"
- [x] Create knobs for each of these standard props and group as design:
-`civicMapStyle`

**Custom**
- [x] Create "Custom" story
- [x] Create knobs for each of these custom props and group as custom:
-`initialLongitude`
-`initialLatitude`
-`initialZoom`
-`initialPitch`
-`height`
-`navigation`
-`onBaseMapClick`

**Examples**
- [x] Rename "With geocoder usage" story to "Example: With Geocoder"
- [x] Create knobs for each of these props:
-`geocoder`
-`geocoderOptions`
-`geocoderOnChange`
-`locationMarker`
-`mapGLOptions`

- [x] Rename "No interactivity" story to "Example: No Interactivity"
- [x] Create knobs for each of these props:
-`mapGLOptions`

- [x] Rename "Use container height" story to "Example: Use Container Height"
- [x] Create knobs for each of these props:
-`useContainerHeight`

**Notes**
- [x] Add Base Map Notes file

## Why
**Why does it need to be done? (Include links to related issues)**
Base Map stories should follow the established CIVIC design system story pattern.

## When
**When does this need to be done?**
Soon

PR #644, opened 2019-07-10
Design System - Base Map

mendozaline:
This PR addresses #631.

- It updates the Base Map standard, custom, and example stories 
- It adds Base Map notes 
- It also fixes some linting errors in the Base Map component


mendozaline:
I updated the default props to include the viewport options and also updated the use container height story to explain the use of the `min-height` property.

PR #658, opened 2019-07-16
Feature/slow river

Dianna:
I slowed down the orb velocity. I went ahead and fixed up the screen ratio some as well since most orbs were offscreen. It isn't totally fixed though. I can prevent orbs from going out _over_ the interface, but not out _below_ the interface and I was confused enough I figured I should spend my time on something else. For the ratios, I only considered laptop / desktop sized and touchscreen sized.

How things are moving:
ratios-interface

A better look at the screen ratio
Screen Shot 2019-07-15 at 4 54 06 PM


packages/project-disaster-trail/src/components/Game/OrbManager.js
```diff
@@ -16,13 +16,16 @@ const OrbManager = ({ count = 10 }) => {
...
     model = {};
     model.x = Math.random() * window.innerWidth;
-    model.y = Math.random() * window.innerHeight;
+    model.y =
+      Math.random() * (ratios.interfaceHeight - ratios.orbSize) +
+      ratios.orbSize;
+
     model.velocity = {
-      x: (Math.random() - 2) * 10,
+      x: (Math.random() - 2) * 2,
```
jaronheard: could make velocity a prop, would be nice for tweaking things

PR #669, opened 2019-07-22
Story Cards v2 - add story, add data checker

jaronheard:
Working on the following efforts for Story Cards:
- Add a story showing Layout options for CivicCard
- Added a data checker to CivicCard to validate cardMeta is the right object shape


packages/2019-template/src/components/DemoCard/index.js
```diff
@@ -33,6 +33,7 @@ class DemoCard extends Component {
 }
 
 DemoCard.displayName = "DemoCard";
+DemoCard.tags = demoCardMeta().tags;
```
jaronheard: There's got to be a better way to pass these tags to the card registry without adding this line every time. @BJHunnicutt any ideas?

jaronheard: @Dianna if you have any ideas on this, would be interested in your thoughts

jaronheard: Actually, this may be fine. Having this be explicit lets us comment this out until we're ready to expose cards through the tag system

jaronheard:
@Dianna note to pull out the tag related commits into their own PR

5595a74
e0ecdce

Dianna:
The changes in 5595a74 are already changed in the master branch, but I've cherry-picked e0ecdce into #734

packages/component-library/stories/CivicCard.story.js
```diff
@@ -0,0 +1,335 @@
...
+  metadata: (
+    <p>
+      <em>
+        Note: This dataset documentation is for demonstration purposes only and
+        is not peer-reviewed.
+      </em>
+    </p>
+  ),
+  // metadataQA likely to take metadata API endpoint in the near future
+  metadataQA: [
```
Dianna: When running package 2018, the `TemplateCard` is now erroring out because it doesn't have `metadataQA`, is this property mandatory. If so, we should update the metadata for that card. If not, we should make sure that optional properties aren't interpreted as required.

Dianna: That being said, we should add an example of this to the README

ghost (changes requested):
I've reverted some commits and added them to #734. I've also fixed some small issues. There's one last issue I'm not sure how you want resolved. Let me know and I can do it or you're welcome to jump in.

ghost (approved):
This looks great! Only one comment. About the commit e942622 which removes the "unused files" for the template API: I'm fine leaving them out; however, they are imported files required by other templates though, so that would lead to an error if the package didn't already have them.

Dianna:
Never mind! Those are only used in the template file card. So we are totally good to go.

PR #670, opened 2019-07-22
Change default civic-card layout to full, simplify 2019-template index

jaronheard:
A couple of simple simplifying tweaks to `CivicCard` and our `2019-template`


ghost (approved):
Really great work here

PR #671, opened 2019-07-22
Add internal packages to 2019-template package

jaronheard:
This makes ESLint happy -- but whoever reviews this please double check that this makes sense.

Maybe `@hackoregon/dev-server` should be a dev dependency?

### DIFF
--- a/packages/2019-template/package.json
+++ b/packages/2019-template/package.json
@@ -33,7 +33,6 @@
   },
   "dependencies": {
     "@hackoregon/component-library": "^3.0.0",
-    "@hackoregon/dev-server": "^3.0.0",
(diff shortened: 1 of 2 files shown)


packages/2019-template/package.json
```diff
@@ -33,6 +33,9 @@
   },
   "dependencies": {
     "@hackoregon/component-library": "^3.0.0",
+    "@hackoregon/dev-server": "^3.0.0",
```
Dianna: should be in devDependencies

ghost (changes requested):

ghost (approved):
LGTM 👍

PR #673, opened 2019-07-23
Feature range slider

Foadm:
Added a multiple knob Range RCslider.
Screen Shot 2019-07-23 at 8 42 37 AM
Screen Shot 2019-07-23 at 8 42 46 AM


packages/component-library/src/Slider/Slider.js
```diff
@@ -1,20 +1,22 @@
...
   min,
   max,
   onChange,
   showStepMarkers,
   step,
   value,
-  SliderComponent
+  SliderComponent,
+  sliderType,
+  defaultValue
```
Dianna: Can you remove the `sliderType` and `defaultValue` props from this file? They don't appear to be used anywhere and we want to avoid dead code.

PR #678, opened 2019-07-26
Update/multi map storybook

mendozaline:
This updates the storybook stories for the Multi-Layer Map to fix some issues encountered by initial users.
- First, it groups everything into just 2 tabs: Design and Data. This should make it simpler to navigate for people unfamiliar with storybook. 
- It updates the notes file and corrects some typos 
- It also adds a new knob that allows users to update where the features array is for their particular data endpoint (i.e., `data.features` vs. `data.results.features` vs. `data.slide_data.features`).
- In addition to rgb color arrays (e.g., `[255,0,0]`), users can now enter hex color codes (e.g., `"#ff0000"`) into the storybook color range knob

### DIFF
--- a/packages/component-library/src/MultiLayerMap/createLayers.js
+++ b/packages/component-library/src/MultiLayerMap/createLayers.js
@@ -1,5 +1,4 @@
 /* eslint-disable no-nested-ternary */
-/* eslint-disable import/no-extraneous-dependencies */
 import {
   scaleQuantize,
   scaleThreshold,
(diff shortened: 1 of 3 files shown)


packages/component-library/src/MultiLayerMap/createLayers.js
```diff
@@ -1,14 +1,37 @@
 /* eslint-disable no-nested-ternary */
+/* eslint-disable import/no-extraneous-dependencies */
```
jaronheard: I'm trying to be more careful about this error -- does this mean that d3 isn't a component-library dependency, and it's just being hidden because it's imported elsewhere? If that's the case, let's add it as a dependency.

mendozaline: I removed that line since I'm not sure why I added it in the first place. The ESLint hook didn't show any errors when I tried to commit the change.

PR #679, opened 2019-07-26
Feature/task state

Dianna:
Adds `tasks` to the application's state. Includes some example tasks, environments where tasks are applicable, and enables adding tasks (like we would on a vote) and changing the environment (like we would at the end of the game)

You can check out the tasks by clicking to the tasks chapter OR you can go to `project-disaster-trail/src/constants/chapters.js` and update CHAPTERS.2.enabled to `false` and CHAPTERS.6.enabled to `true` if you'll be refreshing much.

tasks

### DIFF
--- a/packages/project-disaster-trail/src/components/Game/TaskScreen.js
+++ b/packages/project-disaster-trail/src/components/Game/TaskScreen.js
@@ -1,0 +1,121 @@
+import React from "react";
+import { PropTypes } from "prop-types";
+import { connect } from "react-redux";
+import { map } from "lodash";
+import {
+  getTaskOrder,
+  getActiveTask,
+  getActiveEnvironment,
+  getTasksForEnvironment,
+  doNextTask,
+  changeEnvironment
+} from "../../state/tasks";
+
+const TaskScreen = ({
+  taskOrder,
+  activeTask,
+  activeEnvironment,
+  tasksForEnvironment,
+  addTask,
+  updateEnvironment
+}) => {
+  const possibleTasks = [];
+  map(tasksForEnvironment[activeEnvironment], tasks => {
+    for (let i = 0; i < tasks.length; i += 1) {
+      const taskOption = tasks[i];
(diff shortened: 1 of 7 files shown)


packages/project-disaster-trail/src/components/Game/TaskScreen.js
```diff
@@ -0,0 +1,121 @@
...
+
+const TaskScreen = ({
+  taskOrder,
+  activeTask,
+  activeEnvironment,
+  tasksForEnvironment,
+  addTask,
+  updateEnvironment
+}) => {
+  const possibleTasks = [];
```
jaronheard: This is really just preference, but I'd write this more simply using native Array.map:
```js
const possibleTasks = tasksForEnvironment[activeEnvironment].map(task => (
  <button
    key={task}
    type="button"
    onClick={() => {
      addTask(task);
    }}
  >
    {task}
  </button>
));
```

packages/project-disaster-trail/src/components/Game/TaskScreen.js
```diff
@@ -0,0 +1,121 @@
...
+          }}
+        >
+          {taskOption}
+        </button>
+      );
+    }
+  });
+
+  const environments = Object.keys(tasksForEnvironment);
+  const envButtons = [];
```
jaronheard: Similarly to above, would simplify using native Array.map

packages/project-disaster-trail/src/components/Game/TaskScreen.js
```diff
@@ -0,0 +1,121 @@
+import React from "react";
+import { PropTypes } from "prop-types";
+import { connect } from "react-redux";
+import { map } from "lodash";
```
jaronheard: To reduce overhead for other developers, use native ES6 Array.map instead.

Dianna:
All issues have been addressed

PR #645, opened 2019-07-11
Bar chart sorting final

corydm92:
Fixed, from branch issue-#574-sort-order. Used git cherry-pick to isolate commits into one branch.

PR #680, opened 2019-07-27
Issue #574 sort order

corydm92:
Tests changed to reflect compare method


jaronheard:
Hey @corydm92 -- It looks like this branch has other unrelated changes. Can you cherry-pick 1db48df to the `bar-chart-sorting-final` branch, and then push that branch? 

Then we can merge #645.

That process would look something like this:
```bash
git checkout bar-chart-sorting-final
git cherry-pick 1db48df
/* run tests and make sure tests pass */
git push origin bar-chart-sorting-final
```

PR #682, opened 2019-07-30
Update .gitignore to include package-lock.json

jaronheard:
Added package-lock.json to .gitignore so that it can't accidentally be committed.

If you have a package-lock.json file show up in your repo, it means that you've ran `npm` instead of `yarn` -- it's happened to me!


ghost (approved):
Tested to double check that it does indeed block adding package.json... and it does! Woohoo! I deleted the remaining `package-lock.json` as well

PR #686, opened 2019-07-31
Add emotion/core to 2019 packages

jaronheard:


ghost (approved):
Approving with a question. `@emotion/core` isn't actually being used in any of these projects. I'm no webpack pro, but I _think_ that's ok. From what I remember webpack doesn't include packages in the bundle unless they are actually used, but I wanted to double check that with you.

PR #687, opened 2019-07-31
Add redux-devtools-extension to 2018 & mockwrapper

jaronheard:
Our previous method of applying the redux-dev-tools extension wasn't working. It works now in project packages and year packages!

image

### DIFF
--- a/packages/2018/package.json
+++ b/packages/2018/package.json
@@ -33,7 +33,7 @@
     "@hackoregon/2018-transportation-systems": "^3.0.0",
     "@hackoregon/2019-housing": "^1.0.0",
     "@hackoregon/2019-template": "^1.0.0",
-    "@hackoregon/2019-transportation": "^1.0.0",    
+    "@hackoregon/2019-transportation": "^1.0.0",
     "@hackoregon/civic-sandbox": "^3.0.0",
     "@hackoregon/component-library": "^3.0.0",
     "chalk": "^1.1.3",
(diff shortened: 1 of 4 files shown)


packages/2018/package.json
```diff
@@ -48,6 +48,7 @@
     "react-router-redux": "^4.0.8",
     "react-typekit": "^1.1.3",
     "redux": "^4.0.1",
+    "redux-devtools-extension": "^2.13.8",
```
Dianna: should be in devDependencies

packages/mock-wrapper/package.json
```diff
@@ -29,6 +29,7 @@
     "react-router": "^3.0.0",
     "react-router-redux": "^4.0.8",
     "redux": "^4.0.1",
+    "redux-devtools-extension": "^2.13.8",
```
Dianna: Should be in devDependencies

packages/2018/package.json
Dianna: Nevermind! This is used in the `App.js` file

ghost (approved):
Since the `mock-wrapper` package is only used in local development I've moved the extension to dev-dependencies and eslint-ignored the import. I've left 2018 as is because it sounds like that will actually be part of the production build. I'm not performance expert, and certainly not a build expert, but if the extension is part of the production build that seems like a red flag to me and perhaps requires conditional loading and use

PR #695, opened 2019-08-02
Remove unused dependencies from component-library

### LINKED ISSUE #689: Extraneous component library dependencies causing issues in open-elections
## Description

We've got unused dependencies that may be causing issues in the `open-elections` project in the component-library - an example would be `react-router`-- the component-library isn't doing any router.

jaronheard:
Resolves #689

- Only addresses dependencies, not devDependencies
- We are using react-router for the CivicStoryCard component. If having this as a dependency causes issues for other projects that rely on the component library, then we should look at options to remove it


Dianna:
depcheck: https://www.npmjs.com/package/depcheck
and
npm-check: https://github.com/dylang/npm-check
may be some options worth checking out for long term dependency management.

packages/component-library/package.json
```diff
@@ -74,36 +75,26 @@
   },
   "dependencies": {
     "@material-ui/core": "^4.1.3",
-    "chai-enzyme": "^1.0.0-beta.1",
```
Dianna: This package is being used in `mocha.conf.js` which is used for testing
https://github.com/hackoregon/civic/blob/master/packages/component-library/mocha.conf.js#L50

Dianna: Ah ha! This is added to `devDependencies`. Nothing to see here.

PR #696, opened 2019-08-02
Update axios to latest to resolve security alerts

jaronheard:
See https://github.com/hackoregon/civic/network/alerts


Dianna:
@jaronheard, I get a 404 when I try to go to that link

PR #688, opened 2019-07-31
WIP - Transportation Storycard 001

jaronheard:

PR #718, opened 2019-08-03
Feature/template api card

ghost:
New card = TemplateAPICard
Rename TemplateCard to TemplateFileCard

Have a demo card that can be replicated in the 2019-template projects
- Start with the card in https://github.com/hackoregon/civic/pull/688/ for reduxful pattern
- Add a new card to the 2019-template project that duplicates:
  - http://civicplatform.org/cards/plateau-in-ridership
  - Any new parts of the storycard structure can be based on template-card 
  - Should have the same functionality and structure, but instead of loading from a file, use reduxful loading pattern
  - Should be as simple as possible for developer


packages/2019-template/src/components/TemplateAPICard/index.js
```diff
@@ -0,0 +1,62 @@
+import React, { useEffect } from "react";
+import PropTypes from "prop-types";
+import { connect } from "react-redux";
+import { isLoaded } from "reduxful";
+import { resourceShape } from "reduxful/react-addons";
+import { CivicStoryCard } from "@hackoregon/component-library";
+
+import api from "../../state/plateau-in-ridership/api";
+
+const TemplateAPICard = ({ init, data }) => {
```
jaronheard: Get out of here, class components! ;)

packages/2019-template/src/state/plateau-in-ridership/api.js
```diff
@@ -0,0 +1,20 @@
...
+  }
+};
+
+const plateauInRidershipAPI = new Reduxful(
+  "plateauInRidershipAPI",
+  apiDesc,
+  apiConfig
+);
+
+export default plateauInRidershipAPI;
```
jaronheard: Oh my gosh this pattern is so much cleaner.

An actual question:
I was going to suggest renaming the folder containing this api to template-api-card -- so all of the state for a card lives in a corresponding folder with the same name as the component.

However, if you have an API that you want to use across multiple cards, then keeping a name that corresponded to what the API did could make sense.

How much should live in one Reduxful file?

Dianna: Right now its clean enough that I actually like having it all in one file... I'm down with the renaming. It's much clearer then what corresponds to what.

BJHunnicutt:
@Dianna & @jaronheard -- I just pushed up some changes that add hygen generators for an api story card & a local-data story card. I **_think_** I got everything, but it would be great if someone had a second to double check them! Just run `yarn card` or `yarn card:local-data` & enter the package & card names 🤞

_templates/api-story-card/with-prompt/src/components/App/index-import.ejs.t
```diff
@@ -0,0 +1,7 @@
+---
+to: packages/<%=package%>/src/components/App/index.js
+inject: true
+after: \nimport(?![\s\S]*\nimport).*
```
BJHunnicutt: This regex will inject this after the the last occurrence of a new line that begins with "import" (i.e. right after the other import statements) I tried to add a comment in here clarifying that, but I could not figure out how to add comments to these files without breaking something...

_templates/api-story-card/with-prompt/src/state/api.ejs.t
```diff
@@ -0,0 +1,20 @@
+---
+to: packages/<%=package%>/src/state/<%=slug%>-data/api.js
+---
+import Reduxful from "reduxful";
+import requestAdapter from "../request-adapter";
+
+const apiConfig = { requestAdapter };
+
+// FIXME: getRidershipOverTime should be variable (or more clearly mock data)
```
BJHunnicutt: I wasn't sure if we wanted to just always use this as the dummy data, but if so, would it make sense to name it something more obviously fake? Maybe something like: `getMockRidershipData`?

_templates/api-story-card/with-prompt/src/components/App/index-card.ejs.t
```diff
@@ -0,0 +1,9 @@
+---
+to: packages/<%=package%>/src/components/App/index.js
+inject: true
+before: '</PageLayout>'
+skip_if: <<%=StoryCardName%>
```
Dianna: I'm not 100% or even 40% familiar with hygen, but I feel like `<<%=StoryCardName%>` isn't the syntax that should be here. Should the opening carat be removed or a final `/>` be added?

Dianna: Ahh. Ok. That is it. I see in the addition below that if `<ComponentName` isn't present then we're adding the section below.

_templates/api-story-card/with-prompt/src/components/App/index-import.ejs.t
Dianna: I think [this solution may work](https://github.com/tj/ejs/issues/76#issuecomment-14104705)

_templates/api-story-card/with-prompt/src/index-import.ejs.t
```diff
@@ -0,0 +1,7 @@
+---
+to: packages/<%=package%>/src/index.js
+inject: true
+after: \nimport(?![\s\S]*\nimport).*
```
Dianna: TODO: add comment for this line

package.json
```diff
@@ -99,6 +100,7 @@
     "file-loader": "^3.0.1",
     "html-webpack-plugin": "^3.2.0",
     "husky": "^1.3.1",
+    "hygen": "^4.0.7",
```
Dianna: It looks like you didn't add the lockfile so I get an error when I run `yarn bootstrap`
```bash
dna Civic: (feature/template-api-card)$ yarn bootstrap
yarn run v1.16.0
$ yarn install --frozen-lockfile && yarn --cwd ./packages/component-library/ build && yarn --cwd ./packages/mock-wrapper/ build
[1/4] 🔍  Resolving packages...
error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

jaronheard (commented):
Just a comment about some dependencies and the `request-adapter.js` file

package.json
Dianna: I added the lockfile which resolves this error.

Dianna: @BJHunnicutt, make sure to add the `yarn.lock` that's generated whenever you add something to `package.json`. That file makes sure that every developer is using the correct version of dependencies and that those dependencies can be effectively symlinked by yarn workspaces

packages/2019-template/src/state/request-adapter.js
```diff
@@ -0,0 +1,4 @@
+import fetch from "cross-fetch";
+import { makeFetchAdapter } from "reduxful";
+
+export default makeFetchAdapter(fetch);
```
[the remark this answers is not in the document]
Dianna: I noticed this as well. I'm agreed on adding the `request-adapter.js` file if it doesn't already exist.

I was thinking we could just inject the dependencies using the same hygen template techniques from this PR. Generally speaking, I don't like adding dead code, which is why I prefer injecting if not present rather than assuming that a new package will be making cards... but then again only the disaster game so far has not included cards. I'll add this to my final summary.

_templates/local-data-story-card/with-prompt/src/state/actions.ejs.t
```diff
@@ -0,0 +1,18 @@
+---
+to: packages/<%=package%>/src/state/<%=slug%>-data/actions.js
+---
+import importAdapter from "../import-adapter";
+import actionEmitter from "../import-adapter-action-emitter";
```
Dianna: Note to self: what to do if dev hasn't created these files yet?

_templates/local-data-story-card/with-prompt/src/components/index.ejs.t
```diff
@@ -0,0 +1,55 @@
+---
+to: packages/<%=package%>/src/components/<%=StoryCardName%>/index.js
+---
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import { connect } from "react-redux";
+import { CivicCard } from "@hackoregon/component-library";
+
+import <%=StoryCardName%>Meta from "./<%=storyCardName%>Meta";
+import { fetch<%=StoryCardName%>Data } from "../../state/<%=slug%>/actions";
```
Dianna: In `_templates/local-data-story-card/with-prompt/src/state/actions.ejs.t` the path where `actions.js` is created is would actually be 
`../../state/<%=slug%>-data/actions;`
not:
`../../state/<%=slug%>/actions;`

_templates/local-data-story-card/with-prompt/src/components/index.ejs.t
```diff
@@ -0,0 +1,55 @@
...
+import {
+  is<%=StoryCardName%>DataPending,
+  get<%=StoryCardName%>Data
+} from "../../state/<%=slug%>/selectors";
```
Dianna: Same file path issue here

_templates/local-data-story-card/with-prompt/src/state/index.ejs.t
```diff
@@ -0,0 +1,12 @@
+---
+to: packages/<%=package%>/src/state/<%=slug%>-data/index.js
+---
+import importReducer from "../import-adapter-reducer";
```
Dianna: Note to self: What if file doesn't yet exist?

_templates/local-data-story-card/with-prompt/src/index-import.ejs.t
```diff
@@ -0,0 +1,7 @@
+---
+to: packages/<%=package%>/src/index.js
+inject: true
+after: \nimport(?![\s\S]*\nimport).*
```
Dianna: TODO: add explanatory comment

_templates/local-data-story-card/with-prompt/src/state/actions.ejs.t
```diff
@@ -0,0 +1,18 @@
+---
+to: packages/<%=package%>/src/state/<%=slug%>-data/actions.js
```
Dianna: See import issue comment

_templates/local-data-story-card/with-prompt/src/state/index-import.ejs.t
```diff
@@ -0,0 +1,7 @@
+---
+to: packages/<%=package%>/src/state/index.js
+inject: true
+after: \nimport(?![\s\S]*\nimport).*
```
Dianna: TODO: exp. comment

_templates/local-data-story-card/with-prompt/src/state/index.ejs.t
```diff
@@ -0,0 +1,12 @@
+---
+to: packages/<%=package%>/src/state/<%=slug%>-data/index.js
```
Dianna: See import issue comment

_templates/local-data-story-card/with-prompt/src/state/selectors.ejs.t
```diff
@@ -0,0 +1,20 @@
+---
+to: packages/<%=package%>/src/state/<%=slug%>-data/selectors.js
```
Dianna: See import issue comment

_templates/api-story-card/with-prompt/src/state/index-import.ejs.t
```diff
@@ -0,0 +1,7 @@
+---
+to: packages/<%=package%>/src/state/index.js
+inject: true
+after: \nimport(?![\s\S]*\nimport).*
```
Dianna: TODO: exp. comment

_templates/local-data-story-card/with-prompt/src/components/App/index-import.ejs.t
```diff
@@ -0,0 +1,7 @@
+---
+to: packages/<%=package%>/src/components/App/index.js
+inject: true
+after: \nimport(?![\s\S]*\nimport).*
```
Dianna: TODO: exp. comment

Dianna:
@BJHunnicutt, great work on this!

@jaronheard, I have some remaining comments / questions / recommendations:
* Is the data we are importing a good example of "standard" data and how it would be accessed? If so, I'm wondering about Jeannie's suggestion to give the data a more obvious mock / example-esque name. Obviously it would rapidly get subbed out, but it does make it more obvious what came from the template. Thoughts?
* Let's add the `request-adapter.js` file.
* What should we do about dependencies required by the created files? I see 3 routes, and I like the 1st best:
  1) We inject them into `package.json` using standard hygen procedure. This is the most auto-magical solution.
  2) We make it a part of the standard package generator's `package.json`. I don't like this solution because it introduces dead code.
  3) We add documentation with a line of code to add all the dependencies. This gives the developer a chance to opt out, but I don't really see the point in opting out.
* There's issues with the name the generator gives the state directory vs what the `app/index.js` attempts to import. Do we want the directory with our without the word `-data` at the end?

Dianna:
I forgot to add:
What do we want to do if the dev has not yet created a `state/import-adapter.js` or `state/import-adapter-action-emitter.js` file? We could make the generator add these if they don't exist. They look template-able.

jaronheard:
@Dianna @BJHunnicutt here are my thoughts!

> Is the data we are importing a good example of "standard" data and how it would be accessed? If so, I'm wondering about Jeannie's suggestion to give the data a more obvious mock / example-esque name. Obviously it would rapidly get subbed out, but it does make it more obvious what came from the template. Thoughts? 

Yep, let's rename it to getMockRidershipData.

> Let's add the request-adapter.js file.

Yep!

> What should we do about dependencies required by the created files? I see 3 routes, and I like the 1st best:
> We inject them into package.json using standard hygen procedure. This is the most auto-magical solution.

I also like injecting the dependency into the `package.json`, because magic✨. If we do this, we may want to either run `yarn bootstrap` after we run `yarn card`, or at least warn people that they need to do it

> There's issues with the name the generator gives the state directory vs what the app/index.js attempts to import. Do we want the directory with our without the word -data at the end?

Don't have a strong opinion, but it one is needed I'll vote no -data.

> What do we want to do if the dev has not yet created a state/import-adapter.js or state/import-adapter-action-emitter.js file? We could make the generator add these if they don't exist. They look template-able.

Let's add it to the generator!

Dianna:
I've added the changes discussed above and this is ready for review!

jaronheard (approved):
This is great! Looks ready to merge in.

A couple of ideas for next step improvements:
- Replace the ice cream ipsum with more relevant example text
- Demonstrate or comment where to transform and/or filter API response to format needed for data visualization if necessary
- Include data visualization parameters when added to CivicCard

Issue #703, opened 2019-08-02
Task chapter: transitions


Dianna:
The Task chapter, which is after the quake chapter, should follow this sequence from start to finish:

1) Task Solving Screen (#701): Save yourself tasks in sequence
2) The upper screen lifts (excluding the match-lock interface and duration bar) (#727)
3) The map is revealed with disaster icons (#698)
4) Upon choosing, the map pans & zooms to chosen task (#698)
5) The Task Solving Screen (#701) drops from the top: chosen task
6) Repeat 2-5 until the time for the Task chapter runs out

### Requirements
POC: proof of concept

- [x] Create any screen component that is not yet implemented (no need to implement it's whole design and feel free to reshape the size and layout of interim POC screens)
- [x] Create the sequence from one screen state to the next
  - [x] For panning / zooming, if the map isn't implemented, for POC just make sure a function is implemented which will make the text "Pan and zoom" show on the screen before task solving screen drops down.
- [x] Add the lifting and dropping animations
- [x] Create issue for summary screen which is TBD

Dianna:
Completed looooong ago

PR #722, opened 2019-08-04
Feature/task flow

Dianna:
* Changes the Task chapter flow to match team decision to first solve all save yourself tasks before showing the task choosing screen (see #703)
* Abstracts the future map component from the solve task screen. This makes the method, state, and props separation more intuitive and less cluttered — especially for future development.


Dianna:
@jaronheard, fixed that bug I found Saturday that was preventing the merge. That bug being, during the save yourself phase, if the task timed out it didn't reset the timer. Now ready for review.

Issue #737, opened 2019-08-08
Running yarn card for the project-disaster-trail project results in :boom:


DingoEatingFuzz:
## Description

Running the hot new command on the block, `yarn card` providing `project-disaster-trail` as the package results in generated code that is syntactically invalid.

### Steps to reproduce

1. From a clean master branch, after running `yarn bootstrap`
2. Run `yarn card`
3. Use `project-disaster-trail` as the package to put the card in

### Expected result

Either a syntactically valid successful card generation or an error message saying that a card couldn't be made (and a clean git index, since the operation was unsuccessful).

### Actual result

`packages/project-disaster-trail/src/index.js` has a syntax error

The syntax error appears to be a card registry entry stuffed at the top of the file.

```js
  {
    slug: "stay-safe-friends",
    component: StaySafeFriends
  },
import App from "./components/App";
import Routes from "./routes";
import Reducers from "./state";
import StaySafeFriends from './components/StaySafeFriends'

export { App, Routes, Reducers };
```

### Environment

Typical checkout on macos.

DingoEatingFuzz:
Fixed by #739

PR #739, opened 2019-08-08
Introduce a directive for skipping hygen rules

DingoEatingFuzz:
Since hygen only supports regular expressions for skipping injections, which is particularly bad for matching the absence of text.

I hereby declare a new `// hygen:skip` pattern to make it easier to write hygen skip rules.

The origin of this work was #737, which details how an errant insertion can cause a syntax error.


ghost (approved):
Nailed it ⚰️

PR #743, opened 2019-08-08
Choropleth and Circle Map

mendozaline:
The [data visualization team](https://docs.google.com/document/d/1wCFSziHDl84bBe_SUsqThDccuY9VHL_mhCl8dKFJLNc/edit) requested a Choropleth Map with the ability to overlay circles in the center of each polygon.

This PR accomplishes that by adding two new props to the MapOverlay component: `circleFieldName` and  `radiusScale`. If a string is provided for the `circleFieldName` prop, the component will use turf to calculate the center of each polygon and size it according to that data property value.

Choropleth Circle Map


packages/component-library/stories/MapOverlay.story.js
```diff
@@ -260,6 +280,100 @@ export default () =>
...
+                      return f.geometry.type === "Point"
+                        ? circleFillColor
+                        : f.geometry.type === "Polygon" ||
+                          f.geometry.type === "MultiPolygon"
+                        ? polygonColorScale(f.properties[polygonFieldName])
+                        : [0, 0, 0];
+                    }}
+                    radiusScale={radiusScale}
+                    circleFieldName={circleFieldName}
+                    autoHighlight
```
jaronheard: `autoHighlight` feels a bit funky without a tooltip, so might remove this line. Makes me curious to see a tooltip implementation, though!

Issue #717, opened 2019-08-03
Add `yarn test:watch` in all project packages and fix related warning


jaronheard:
Some project packages have a `test:watch` script, others do not. All project packages should have it.

The script should be
`"test:watch": "yarn run test -w"`, not `"test:watch": "yarn run test -- -w",` to fix the warning below 

```warning From Yarn 1.0 onwards, scripts don't require "--" for options to be forwarded. In a future version, any explicit "--" will be forwarded as-is to the scripts.```

jaronheard:
#751 closed

PR #751, opened 2019-08-10
Add test:watch to all project packages

jaronheard:
Resolves issue #717


lerna.json
```diff
@@ -1,7 +1,7 @@
 {
   "version": "3.0.0",
   "useWorkspaces": true,
-  "npmClient": "npm",
+  "npmClient": "yarn",
```
jaronheard: Need to see if this has other unexpected impacts

PR #752, opened 2019-08-11
Update hygen template to use standardized redux store names

### LINKED ISSUE #748: Cards generated from yarn card not connecting to Redux state in 2018 package
## Description

Run `yarn card` in a project package, then run `yarn start` in the 2018 package. You can see through observing Redux actions that the API call is successful, but data is not passed to the card from Redux, so the chart never loads.

### Steps to reproduce

See above

### Expected result

Data should be passed to the card from Redux as props, and chart should display once API returns data.

### Actual result

API returns data, data for card remains undefined

jaronheard:
Resolves #748 

Also standardizes the way that package names are converted to properties of the Redux Store so that `2019-template` becomes `package2019Template` and `package<%= h.changeCase.pascalCase(package)%>` in Hygen syntax.

In order to keep consistency, I adjusted the names of the properties of the Redux Store for the other packages used in the 2018 package to match this pattern.

**Before**
image

**After**
image


ghost (approved):
Works well for me locally and code looks nicely organized 👍

PR #753, opened 2019-08-11
Project Disaster Trail TaskMap

DingoEatingFuzz:
First pass at the TaskMap 🎉 

**Done so far:**
  - [x] Add a map
  - [x] Make it all cool and angled (pitch if you want to be technical)
  - [x] Make it non-interactive
  - [x] Place an icon
  - [x] Make it "fly" to coordinates (animate prop)
  - [x] Add mock location coordinates to tasks in the constants file
  - [x] Use `weightedTasks` to provide points on the map with icons mapped to task type
  - [x] Refactor `TaskMap` to take an `activeTask` and always animate to it when it changes
  - [x] Refactor `goToTask` to set the `activeTask` in the local state

I know there are still more things to do, namely make the map "fly" to a coordinate and keep track of icons/tasks. However, I'm fuzzy on the details here, so I'm opening a draft PR with what I have until I get some clarification.

In the event that I disappear or something, at least my work is on GitHub ✨ 

image

@Dianna, working off the existing `TaskMap` component, the intended props are `completedTasks`, which appears to be an array of strings (maybe identifiers?) and `animateMap`, which is a boolean that alludes to panning and zooming.

I'm fuzzy on what the user flows/state transitions here are. My naive understanding of everything makes me want to remodel this as `activeTask` and `completedTasks`, and change tasks to have coordinates/a geojson feature as a property. 

We should probably chat to hash this out.

### DIFF
--- a/packages/component-library/src/BaseMap/BaseMap.js
+++ b/packages/component-library/src/BaseMap/BaseMap.js
@@ -1,6 +1,11 @@
 /* eslint-disable no-nested-ternary */
 import React, { Component } from "react";
-import MapGL, { StaticMap, NavigationControl, Marker } from "react-map-gl";
+import MapGL, {
+  StaticMap,
+  NavigationControl,
+  Marker,
+  FlyToInterpolator
+} from "react-map-gl";
 import Dimensions from "react-dimensions";
 import { css } from "emotion";
 import PropTypes from "prop-types";
@@ -142,7 +147,8 @@ class BaseMap extends Component {
       mapboxLayerType,
       mapboxLayerOptions,
       mapboxLayerId,
-      locationMarkerCoord
+      locationMarkerCoord,
+      animate
     } = this.props;
 
     viewport.width = containerWidth || 500;
@@ -184,12 +190,19 @@ class BaseMap extends Component {
         : CIVIC_LIGHT;
 
     const MapComponent = isInteractive ? MapGL : StaticMap;
+    const animationProps = !animate
+      ? {}
+      : {
+          transitionDuration: 1000,
+          transitionInterpolator: new FlyToInterpolator()
+        };
 
     return (
       <div className={mapWrapper}>
         <MapComponent
           className="MapGL"
           {...viewport}
+          {...animationProps}
           mapStyle={baseMapboxStyleURL}
           mapboxApiAccessToken={mapboxToken}
           onViewportChange={newViewport => this.onViewportChange(newViewport)}
@@ -263,6 +276,7 @@ BaseMap.propTypes = {
   children: PropTypes.node,
   useContainerHeight: PropTypes.bool,
   updateViewport: PropTypes.bool,
+  animate: PropTypes.bool,
   onBaseMapClick: PropTypes.func,
   mapboxDataId: PropTypes.string,
   mapboxData: PropTypes.shape({
@@ -280,6 +294,7 @@ BaseMap.defaultProps = {
   geocoder: false,
   useContainerHeight: false,
   updateViewport: true,
+  animate: false,
   initialLongitude: -122.6765,
   initialLatitude: 45.5231,
   initialZoom: 9.5,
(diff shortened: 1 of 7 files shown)


DingoEatingFuzz:
Alright, I deem this Good Enough™ to merge.

@mendozaline, do you mind giving the changes to `BaseMap` a review?

packages/component-library/src/BaseMap/BaseMap.js
```diff
@@ -182,18 +188,32 @@ class BaseMap extends Component {
         ? CIVIC_DARK
         : CIVIC_LIGHT;
 
+    const animationProps = !animate
+      ? {}
+      : {
+          transitionDuration: 1000,
```
Dianna: Ideally we should be able to modify the duration, I think.

packages/component-library/src/BaseMap/BaseMap.js
```diff
@@ -182,18 +188,32 @@ class BaseMap extends Component {
...
           ref={this.mapRef}
           {...mapGLOptions}
           onClick={onBaseMapClick}
           onLoad={onMapLoad}
+          scrollZoom={isInteractive}
+          dragPan={isInteractive}
+          dragRotate={isInteractive}
+          doubleClickZoom={isInteractive}
+          touchZoom={isInteractive}
+          touchRotate={isInteractive}
```
Dianna: @mendozaline, I can see in this [example](https://hackoregon.github.io/civic/?path=/story/component-lib-maps-base-map--example-no-interactivity) where `mapGLOptions` is used to prevent interaction. If this can be set 2 ways with props I think that will be confusing for future developers. That being said. I've been snooping around and I can't seem to find any documentation that defines what can be in `mapGLOptions` so I'm not entirely clear on what that prop can do.

packages/component-library/src/BaseMap/BaseMap.js
DingoEatingFuzz: Good call.

DingoEatingFuzz:
Thank you for the review @mendozaline! Really good feedback, I knew you were the person to review map stuff 😄 

> 1). The "fly to" feature works well. I would only suggest that the knob and prop for civicMapStyle be removed from this story so that the "fly to" buttons appear first and are not hidden in the "Other" tab.

Done! Also added an `animationDuration` slider per @Dianna's comment.

> 2). Regarding the isInteractive prop: is there a reason why the mapGLOptions prop couldn't be used to make the map non-interactive as demonstrated in the "Example: No Interactivity" story? This seems to duplicate what can already be done with an existing prop.

Nope! I just didn't know about the `mapGLOptions` prop or the existing no interactivity story. So I removed my new story and dropped the `isInteractive` prop in favor of using `mapGLOptions` in the disaster packages's `TaskMap`.

I will say that I think there is value in creating some sort of `isInteractive` shorthand property that maps to the corresponding `mapGLOptions` prop, but I'm happy to kick that can down the road.

> 3). Follow up issues should be created for things like keeping the naming consistent by beginning these 2 new stories with "Example:" and adding any new stories and props to the Base Map's notes file. It would benefit everyone to keep component notes as up to date as possible.

More stuff I didn't even know to look for! I renamed the animate to coordinates story to match the "Example: " prefix pattern. I also updated the notes doc with my new example.

packages/component-library/src/BaseMap/BaseMap.js
mendozaline: If you click on the Notes tab next to the Canvas tab in the upper-left corner of the Base Map story, you'll see what options can be set with the `mapGLOptions` prop under the "Example: No Interactivity" section. These are just some of the the most common options we've used before, but it also includes a link to the `react-map-gl` [documentation](https://github.com/uber/react-map-gl/blob/master/docs/components/interactive-map.md) that has even more options that can be set with it.

Screenshot_2019-08-12 Storybook

mendozaline:
> I will say that I think there is value in creating some sort of isInteractive shorthand property that maps to the corresponding mapGLOptions prop, but I'm happy to kick that can down the road.

If we encounter usability issues with the `mapGLOptions` prop, it would make sense to make the non-interactivity feature of the Base Map more explicit and reserve the `mapGLOptions` prop for options that are not as frequently used.

PR #754, opened 2019-08-11
Remove redux logger from 2018 and convenience packages

jaronheard:
Now that #687 is complete, we can use the [Redux Dev Tools](https://github.com/reduxjs/redux-devtools) rather than Redux Logger. 

This will allow us to see other console messages much more clearly.

Note -- left 2017 packages alone for now


ghost (approved):
Good call 👍

PR #755, opened 2019-08-11
Fix chart animation issues and VictoryAnimate PropType warnings

jaronheard:
We were using the animate prop on Victory improperly, and it was clouding the console with errors as well as causing issues with animations not completing. This simple PR solves those issues.

Before:
Screen Recording 2019-08-10 at 9 10 04 PM

After:
Screen Recording 2019-08-10 at 9 08 11 PM


jaronheard:
@DingoEatingFuzz This PR relates to the issue that you filed with 
https://github.com/FormidableLabs/victory/issues/1282 - so I recommend taking a look

Issue #572, opened 2019-05-24
Improve story card links for embedded story cards


jaronheard:
## Description

If you embed a story card, using the method described in #538, you'll encounter issues with all three of the links at the bottom of a story card.
- Source: since this links to another domain, needs to open in a new window to avoid iframe security issues
- Improve:  since this links to another domain, needs to open in a new window to avoid iframe security issues
- Share: this should copy a link, but currently redirects you to the /cards route

### Steps to reproduce

Add this node to another website

```
        <iframe
          id="magnitude-of-sweeps"
          type="text/html"
          width="800"
          height="900"
          src="https://civicplatform.org/cards/magnitude-of-urban-campsite-sweeps/embed/"
          frameborder="0"
        />
```

### Expected result

Source: should open source link in new window
Improve: should open improve link in new window
Share: should copy link

### Actual result

Source: iframe security error
Improve: iframe security error
Share: links to /cards route

PR #756, opened 2019-08-11
fixed issue #572

cosmolightfoot:
…m embedded iframe; Share link now copies to clipboard without linking to card page


packages/component-library/src/CivicStoryCard/CivicStoryFooter.js
```diff
@@ -60,25 +60,29 @@ export default class StoryFooter extends Component {
     this.setState({ copied: true });
   };
 
+  handleEmbedLink = url => {
```
jaronheard: looks like this is dead code

jaronheard:
@cosmolightfoot This looks good to go once the dead code commented on above is removed

PR #780, opened 2019-08-15
Update Icon Map to use Material Design Icons

mendozaline:
This PR updates the Icon Map to use Material Design Icons.

However, I ran into an issue with using svg icons created as a component with `@mdi/react`. Instead, I used the path data from `@mdi/js` to create a data URI for each svg icon and passed that as the url property for the `getIcon` prop.

Since the [disaster resilience API endpoint](https://service.civicpdx.org/disaster-resilience/sandbox/slides/poi/) from last year we used for the icon map is returning an internal server error, I had to improvise. Ideally the `iconSet` object would contain the name of each icon and then within the `getIcon` function use the `d.properties.type` from the GeoJSON to set the url. I had to use numbers and create it randomly.

I opened this PR as a draft because even though this works I'm not sure this couldn't be improved. So I'm open to any suggestions, comments, or questions.

### DIFF
--- a/packages/component-library/package.json
+++ b/packages/component-library/package.json
@@ -76,6 +76,8 @@
   "dependencies": {
     "@emotion/core": "^10.0.15",
     "@material-ui/core": "^4.1.3",
+    "@mdi/js": "^3.9.97",
+    "@mdi/react": "^1.2.1",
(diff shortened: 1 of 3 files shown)


jaronheard:
Closing as an unfinished WIP

PR #790, opened 2019-08-16
Feature/style themes

Dianna:
I tried to split this into separate commits so that it's obvious what is happening and why. Here's roughly what happened:

1) I audited the existing styles and pulled out dead code (I also did this in the very last commit)
2) I created object shaped themes in `/component-library/src/_Themes`. I included the underscore so it was easy to find and distinctive from the other components. This whole directory can be renamed and moved elsewhere potentially. It only uses `assign` from lodash so it would be easy enough to make dependency free if desired (which means we wouldn't need to build it)
3) I made a couple examples where I use emotion's `<ThemeProvider/>` in the `BarChart` and `LineChart` components. In the stories, the "CrazyVisualizationTheme" is provided only to the "standard" stories in storybook, so you can see how the default works as planned
4) In `2019-template/` you can see the crazy theme passed in for the API card and that it's color is different. You can also see emotion's `<Global />` in use to apply branding theming which had originally been in `global.styles.css`. These styles effectively carry over to the `2018/` package on the /2019/template/ route.

### Next Steps
* Test in an older 2017 package **(Update: this package is already broken and this includes the data viz components style. For now, I will avoid testing with this package until after demo day, when we have time to address its issues)**
* Look into Storybook styles in `storyStyles.js` which are currently applied directly from an object in the stories
* Look into Map styles

### Conclusions
I think this will be easy to extend across our packages and components. In regards to the components, it won't require that developers change anything they're already doing (unless they want to add a custom theme). Using the `<ThemeProvider/>` in packages is a simple switch over. We may not be able to use `<ThemeProvider/>` and `<Global/>` everywhere, but I think it will get us by most places and the themes they rely on can still be the source of truth for other styling methods that may be necessary.


jaronheard:
Oops -- I accidentally clicked the ready for review button! This is still draft.

@Dianna I've poked around, tested, and thought about this, and I like the approach here. I'm going to add a few comments, but this is heading the right direction! 👩‍🚀🚀

packages/component-library/src/_Themes/DataVisualizationTheme.js
```diff
@@ -0,0 +1,335 @@
...
+  fontWeight: "bold"
+};
+
+// *
+// * Strokes
+// *
+const strokeDasharray = "10, 5";
+const strokeLinecap = "round";
+const strokeLinejoin = "round";
+
```
jaronheard: Maybe the Victory theme here should be nested under `victory`, since there may be other data visualization libraries in use.

Let say for example we add a Sankey from another library, e.g. `d3-sankey`, that takes a theme in a different format.

Another consideration -- we have a our own bespoke components, one example being GradientScale. What part of the theme object is being used there?

Maybe it looks like this?
```js
{
  victory: { /* victory theme object */ },
  d3-sankey: { /* d3-sankey theme object*/ },
  civic: { /* our own style used for our custom components */ }
}
```

I still like having all of the data visualization styles in one place, defined by constants that are outside of the object or objects that get exported. But also, 🤷‍♂️.

packages/component-library/src/_Themes/BrandTheme.js
```diff
@@ -0,0 +1,120 @@
+// Use with emotion's Global component
+
+/* Add all of this and BRANDING to theme for 2018 */
```
jaronheard: Will something else be added to the 2018 package other than this theme?

packages/2019-template/src/components/TemplateAPICard/TemplateAPICardVisualization.js
```diff
@@ -2,6 +2,7 @@ import React from "react";
 import PropTypes from "prop-types";
 import { resourceShape } from "reduxful/react-addons";
 
+import { CrazyVisualizationTheme } from "@hackoregon/component-library/src/_Themes/index";
```
jaronheard: This theme is 🎨 image

packages/component-library/src/_Themes/DataVisualizationTheme.js
jaronheard: Another consideration: should this be DataVisualizationTheme or separated out into ChartTheme and MapTheme?

I think I like keeping it together, but could go either way.

Dianna: Great points here. I'm leaning towards separating out themes that are formatted for different 3rd party sources (i.e., victory visualizations, map visualizations, etc...) and keeping the civic styles as the SOT for any colors / fonts and such that the others need to use.

packages/component-library/src/_Themes/BrandTheme.js
Dianna: Ah, those were my notes from when we chatted. I need to keep digging through to make sure I've found all the style sources we're using though.

packages/2019-template/src/components/TemplateAPICard/TemplateAPICardVisualization.js
Dianna: 💯

packages/component-library/assets/global.styles.css
```diff
@@ -133,48 +57,42 @@ a:hover {
   margin-bottom: 12px;
 }
 
-h1,
-.h1 {
```
Dianna: @jaronheard, I did a regex search excluding `/h1` and `<h1` and didn't find anything that looked like an h1 class. I repeated this for the others as well so I feel ok about removing this but I just wanted to check that there isn't a use case I'm missing.

packages/component-library/src/_Themes/DataVisualizationTheme.js
```diff
@@ -0,0 +1,335 @@
+import { assign } from "lodash";
```
Dianna: We could turn this into a function that returns an object which would allow us to update only a few properties if we want to. The way I see this working is that an object can be taken as a property and we return the default object (the victory theme) with those changes overriding default properties. Kinda like:

```javascript
return {
  ...VictoryTheme,
  ...CustomTheme
}
```

packages/component-library/assets/global.styles.css
jaronheard: @Dianna I don't know of other use cases that you'd be missing

packages/component-library/src/_Themes/DataVisualizationTheme.js
jaronheard: @Dianna That sounds good!

packages/2019-template/src/components/App/index.js
```diff
@@ -2,6 +2,8 @@ import React from "react";
 
 // eslint-disable-next-line import/no-extraneous-dependencies
 import { cx, css } from "emotion"; // eslint-disable-line emotion/no-vanilla
+import { Global } from "@emotion/core";
+import { BrandTheme } from "@hackoregon/component-library/src/_Themes/index";
```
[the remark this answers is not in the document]
Dianna: That makes sense. I'm not committed to having the themes live in the component-library yet, but for now I've made them 1st class exports 📦

packages/component-library/src/_Themes/Brand/BrandTheme.js
```diff
@@ -0,0 +1,118 @@
...
+  ".Title": {
+    fontSize: "3.57rem",
+    lineHeight: "1.2",
+    fontWeight: "300",
+    fontFamily: "'Rubik', sans-serif",
+    marginBottom: "12px"
+  },
+
+  h1: {
+    // fontSize: '36px',
```
Dianna: It looks like this was meant to cascade before. Can I safely pull these out or is it meant for an older browser?

packages/component-library/src/_Themes/Brand/BrandTypography.js
```diff
@@ -0,0 +1,15 @@
+// TODO: update to em?
```
Dianna: There's basically nothing standardized here... what do we want to make the law of the land in Typography land?

jaronheard: @Dianna Since nobody lives in typography land, you've ascended to the throne by default, so you get to make the laws now 👑

packages/component-library/src/_Themes/Brand/BrandTheme.js
jaronheard: You can pull these out, we're relying on `em` support

packages/component-library/src/_Themes/Brand/BrandTypography.js
Dianna: 👸

Dianna:
The base camps for a couple missions have been set up. Here are some of the next missions I can embark on:
* Apply the `BrandTheme`, which replaces `global.styles.css` across packages with emotion's `Global`
* Abstract all the map visualization colors and refactor storybook so that it is using them

packages/2019-template/src/components/TemplateAPICard/TemplateAPICardVisualization.js
```diff
@@ -2,7 +2,7 @@ import React from "react";
 import PropTypes from "prop-types";
 import { resourceShape } from "reduxful/react-addons";
 
-import { LineChart } from "@hackoregon/component-library";
+import { VictoryCrazyTheme, LineChart } from "@hackoregon/component-library";
```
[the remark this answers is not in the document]
Dianna: I'm keeping it all in storybook for now

packages/2019-template/src/components/DemoCard/DemoCardVisualization.js
```diff
@@ -3,7 +3,11 @@ import PropTypes from "prop-types";
 // eslint-disable-next-line import/no-extraneous-dependencies
 import { css } from "emotion"; // eslint-disable-line emotion/no-vanilla
 
-import { LineChart, Button } from "@hackoregon/component-library";
+import {
+  LineChart,
+  Button,
+  CategoricalColors
```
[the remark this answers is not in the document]
Dianna: Yup. Done.

packages/component-library/src/LineChart/LineChart.js
```diff
@@ -188,7 +193,8 @@ LineChart.propTypes = {
   yLabel: PropTypes.string,
   xNumberFormatter: PropTypes.func,
   yNumberFormatter: PropTypes.func,
-  legendComponent: PropTypes.func
+  legendComponent: PropTypes.func,
+  theme: PropTypes.shape({})
```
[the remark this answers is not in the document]
Dianna: Future TODO

packages/component-library/src/Sandbox/constants.js
```diff
@@ -1,64 +1,9 @@
 /* TODO: Fix linting errors */
 /* eslint-disable */
 import { scaleThreshold } from "d3";
+import { VisualizationColors, CategoricalColors } from "../_Themes/index";
```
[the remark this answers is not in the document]
Dianna: Done 👍

packages/component-library/src/VictoryTheme/CivicVictoryTheme.js
```diff
@@ -1,5 +1,9 @@
 import { assign } from "lodash";
```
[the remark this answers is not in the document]
Dianna: @jaronheard, I'm not quite clear what you mean here

packages/component-library/src/_Themes/VisualizationColors.js
```diff
@@ -0,0 +1,61 @@
...
+    [65, 182, 196, 200],
+    [29, 145, 192, 200],
+    [34, 94, 168, 200],
+    [37, 52, 148, 200],
+    [8, 29, 88, 200]
+  ]
+};
+
+export default {
+  sequential
```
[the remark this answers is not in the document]
Dianna: Done 👍

packages/component-library/stories/BarChart.story.js
```diff
@@ -75,6 +76,7 @@ export default () =>
             yLabel={yLabel}
             xNumberFormatter={x => civicFormat[optionSelectX](x)}
             yNumberFormatter={y => civicFormat[optionSelectY](y)}
+            theme={VictoryCrazyTheme}
```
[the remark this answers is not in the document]
Dianna: 🎨 Next time ;)

.storybook/config.js
```diff
@@ -23,4 +26,12 @@ function loadStories() {
   require("../packages/component-library/stories");
 }
 
+const withGlobal = cb => (
+  <>
+    <Global styles={BrandTheme} />
```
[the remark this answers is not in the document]
Dianna: Future TODO it is!

packages/component-library/stories/index.js
```diff
@@ -79,12 +77,12 @@ storiesOf("Design|Brand", module)
   .add("Logos", () => <LogosStyle />)
   .add("Color Theme", () => <ColorThemeStyle />)
   .add("Typography", () => (
-    <div>
+    <>
```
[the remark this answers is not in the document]
Dianna: Will add as a TODO

PR #791, opened 2019-08-17
Add Card Layout with descriptions

jaronheard:
- Refactored CivicCardLayoutFull semantic HTML structure, split other components to new file, and use Emotion 10.
- Added a new layout (`CivicCardLayoutFullWithDescriptions`) which uses a newly added `cardMetaDescriptions` file to provide an annotated view showing each of the sections of a card, shown below

Screenshot_2019-08-17 Storybook

Especially interested in comments on the descriptions.

### DIFF
--- a/packages/component-library/src/ChartTitle/ChartTitle.js
+++ b/packages/component-library/src/ChartTitle/ChartTitle.js
@@ -29,10 +29,10 @@ const subtitleStyle = css`
 
 const ChartTitle = ({ title, subtitle }) =>
   title || subtitle ? (
-    <div>
-      {title ? <h3 className={titleStyle}>{title}</h3> : null}
-      {subtitle ? <span className={subtitleStyle}>{subtitle}</span> : null}
-    </div>
+    <figcaption>
+      {title ? <h2 className={titleStyle}>{title}</h2> : null}
+      {subtitle ? <h3 className={subtitleStyle}>{subtitle}</h3> : null}
+    </figcaption>
   ) : null;
 
 ChartTitle.propTypes = {
--- a/packages/component-library/src/CivicCard/CivicCardLayoutFull.js
+++ b/packages/component-library/src/CivicCard/CivicCardLayoutFull.js
@@ -37,26 +37,16 @@ const authorPhoto = css`
   cursor: pointer;
 `;
 
-const cardMetaItem = css`
-  border: 1px solid red;
-`;
-
-const platformItem = css`
-  border: 1px solid blue;
-`;
-
 function CivicCardLayoutFull({ isLoading, data, cardMeta }) {
   return (
     <Fragment>
       <article>
         <div css={[sectionMarginSmall, sectionMaxWidthSmall]}>
           <header>
-            <h1 css={cardMetaItem} id="title">
-              {cardMeta.title}
-            </h1>
+            <h1 id="title">{cardMeta.title}</h1>
           </header>
           <hr />
-          <section css={cardMetaItem} id="tags">
+          <section id="tags">
             {cardMeta.tags.map((tag, index) => (
               <Chip tag={tag} index={index} />
             ))}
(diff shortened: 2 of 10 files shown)


packages/component-library/src/CivicCard/CivicCardLayoutFull.js
```diff
@@ -1,13 +1,17 @@
-import React, { Fragment } from "react";
+/** @jsx jsx */
+import { jsx, css } from "@emotion/core";
+import { Fragment } from "react";
```
jaronheard: I wanted to use the `<> </>` Fragment syntax here, but in combination with the `/** @jsx jsx */` emotion syntax, I was having an issue.

If I recall if you `import React from "react"` it will work, but that line requires an `eslint ignore`

packages/component-library/src/CivicCard/cardMetaDescriptions.js
```diff
@@ -0,0 +1,65 @@
...
+      "REMOVE? REPLACED WITH METADATA QA - This section will point to the metadata API endpoint containing responses to the dataset metadata questionnaire"
+  },
+  metadataQA: {
+    title: "About This Data",
+    description:
+      "UNDER CONSTRUCTION - This section will point to the metadata API endpoint containing responses to the dataset metadata questionnaire"
+  },
+  resources: {
+    title: "Links and Resources",
+    description: "This section includes links and resources"
```
Dianna: I think this could be more descriptive. Links and resources to what?

packages/component-library/src/CivicCard/cardMetaDescriptions.js
```diff
@@ -0,0 +1,65 @@
...
+  },
+  analysis: {
+    title: "About This Analysis",
+    description:
+      "This section is a description of the analysis and includes a link to the data science notebooks and repository"
+  },
+  metadata: {
+    title: "About This Data",
+    description:
+      "REMOVE? REPLACED WITH METADATA QA - This section will point to the metadata API endpoint containing responses to the dataset metadata questionnaire"
```
Dianna: Yes? If it's not being used anymore. That could be a separate PR.

ghost (approved):
This looks good. The semantic html is definitely an improvement. Heads up, this PR does cause some critical and severe accessibility violations according to the Storybook checker. Just a couple of tiny improvements I'd recommend.

packages/component-library/src/CivicCard/cardMetaDescriptions.js
jaronheard: Improved this section

packages/component-library/src/CivicCard/cardMetaDescriptions.js
jaronheard: Will remove in a separate PR

packages/component-library/src/CivicCard/CivicCardLayoutFull.js
jaronheard: Ah, I dug into it.

You actually have to import React from "react" for the `<> </>`  syntax to work, because [@babel/plugin-transform-react-jsx](https://babeljs.io/docs/en/babel-plugin-transform-react-jsx) transforms it to React.fragment, so if you don't import React you will get an error.

packages/component-library/src/ChartTitle/ChartTitle.js
```diff
@@ -29,10 +29,10 @@ const subtitleStyle = css`
...
 const ChartTitle = ({ title, subtitle }) =>
   title || subtitle ? (
-    <div>
-      {title ? <h3 className={titleStyle}>{title}</h3> : null}
-      {subtitle ? <span className={subtitleStyle}>{subtitle}</span> : null}
-    </div>
+    <figcaption>
+      {title ? <h2 className={titleStyle}>{title}</h2> : null}
+      {subtitle ? <h3 className={subtitleStyle}>{subtitle}</h3> : null}
+    </figcaption>
```
Dianna: Check if this is proper usage of `figcaption` (which should be wrapped in a `figure`)

packages/component-library/src/CivicCard/CivicCardLayoutFull.js
```diff
@@ -48,7 +49,7 @@ function CivicCardLayoutFull({ isLoading, data, cardMeta }) {
           <hr />
           <section id="tags">
             {cardMeta.tags.map((tag, index) => (
-              <Chip tag={tag} index={index} />
+              <Chip tag={tag} index={index} key={generate()} />
```
Dianna: Nice

ghost (approved):
Looks great after that `figcaption` update 👍

packages/component-library/src/ChartTitle/ChartTitle.js
jaronheard: It wasn't quite right! Thankfully I had [@fstorr's awesome reference](https://fstorr.github.io/hack-oregon-civic-components/demo-card-visualization/) and I think I now have it sorted ... at least for this part.

Next step will be to see if I can capture the legend in the ChartContainer. I'll make an issue about that.

Issue #626, opened 2019-06-26
Update Bar Chart component propType


mendozaline:
### What
**What needs to be done?**
The `error` propType for the Bar Chart component needs to be updated.

- [x] Change `error` propType from `bool` to `string` [here](https://github.com/hackoregon/civic/blob/master/packages/component-library/src/BarChart/BarChart.js#L113)
- [ ] Update the error knob in Bar Chart story from boolean to text [here](https://github.com/hackoregon/civic/blob/master/packages/component-library/stories/BarChart.story.js#L125)
- Either make the initial value for this knob an empty string so it doesn't show an error immediately, only once an error message is passed to the knob/prop
- Or add an "Example: With error" story like the one the [CIVIC Story Card component](https://hackoregon.github.io/civic/?path=/story/component-lib-civic-platform-civic-story-card--with-error) uses

### Why
**Why does it need to be done? (Include links to related issues)**
It produces an incorrect `propType` warning at the moment when the component library is running locally.

### When
**When does this need to be done?**
In the near future

PR #795, opened 2019-08-18
updated proptype on barchart component to string #626

cosmolightfoot:


jaronheard:
@cosmolightfoot There's one other small update to close out #626 -- I checked off the checkbox that has been completed

PR #810, opened 2019-08-20
Feature/centralize styles

ghost:
Adds style using emotion's `Global` to all 2019 packages and removes `global.styles.css`. It also updates the template

### DIFF
--- a/_templates/package-generator/with-prompt/src/components/App/index.ejs.t
+++ b/_templates/package-generator/with-prompt/src/components/App/index.ejs.t
@@ -5,13 +5,13 @@ import React from "react";
 
 // eslint-disable-next-line import/no-extraneous-dependencies
 import { cx, css } from "emotion"; // eslint-disable-line emotion/no-vanilla
+import { Global } from "@emotion/core";
 import {
+  BrandTheme,
   PageLayout,
   PullQuote,
   CivicCardLayoutClassic, // eslint-disable-line
 } from "@hackoregon/component-library";
-
-import "@hackoregon/component-library/assets/global.styles.css";
 
 const sectionMarginSmall = css`
   display: block;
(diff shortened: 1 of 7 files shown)


jaronheard (approved):
Looks great!

Just something I was thinking about while reviewing, it'd be great to have our template using the Emotion 10 syntax to show good example code

```
/** @jsx jsx */ 
import { jsx, css } from "@emotion/core"

/* css prop instead of className prop */
```

_templates/package-generator/with-prompt/src/components/App/index.ejs.t
```diff
@@ -36,44 +35,59 @@ const paragraphStyle = css`
...
+  <Fragment>
     <Global styles={BrandTheme} />
     <PageLayout
       teamTitle="<%=title%>"
       heroTitle="🚧 Site Under Construction 🚜"
       heroSubtitle="Playground for building out the <%=year%> <%=title%> team frontend"
       overlay
     >
-      <section className={cx(sectionBodyHeading, sectionMaxWidthSmall)}>
+      <section css={css`
```
jaronheard: There's actually a nicer way to combine these styles:
`css={[sectionBodyHeading, sectionMaxWidthSmall]}`

[(described here, bottom code block)](https://emotion.sh/docs/composition)

jaronheard (commented):
Just one comment about a slightly cleaner way to compose styles using the `css` prop

_templates/package-generator/with-prompt/src/components/App/index.ejs.t
Dianna: Crikey. I just didn't scroll far enough down. Fixing.

jaronheard (approved):

Issue #777, opened 2019-08-15
Improve `Collapsable` component accessibility


jaronheard:
### What

The Collapsable component in the component library needs an overhaul to be accessible to all users

Here's [@fstorr's write up of the details](
https://fstorr.github.io/hack-oregon-civic-components/components/more-less-link/#sig-struct-dam-heading--more)

### Why

https://hackoregon.github.io/civic/?path=/story/design-ux-style-guide--accessibility

### When

As soon as possible :)

PR #811, opened 2019-08-20
Improve collapsable component accessibility #777

jaronheard:
Per [@fstorr's write-up](https://fstorr.github.io/hack-oregon-civic-components/components/more-less-link), I improved the accessibility of the collapsible component.

My implementation ended up being a bit different than the write-up but I referenced it a lot!

My main change was that I used a button instead of a link. There were a few reasons for this, mostly due to restrictions of working with components:

- We don't know what the content is about unless we take a `prop` to provide that description. Actually, now that I think about it, we could take that prop and make it optional.
- Giving the hidden section an `id` is difficult if there may be multiple collapsible components in one page.

☑️ Keyboard focus matches @fstorr's implementation
☑️ Screen reader sounds pretty good (at least to my ears)
☑️ Passes automated accessibility testing

I was looking into adding custom styled focus indicators, but it looked like the place that it would make sense to do that would be in our global styles, so I'm planning to make a follow-up issue

*Extras*
- Added storybook dependencies to component library
- Added react hooks eslint rule

@fstorr I'd love a review! If you checkout my branch, `yarn bootstrap`, then `yarn storybook` you can find the stories of the Collapsible component there


jaronheard:
I added a `description` prop to make the button description more useful to screen reader users.

Went through a review with @fstorr in person tonight. There's a remaining issue, #823, to make sure that focus is visible

jaronheard:
I reverted the commit that added enzyme across all packages to make for easier review

ghost (approved):
Component works and the accessibility drawer on storybook looks happy with the changes so LGTM 👍

PR #816, opened 2019-08-21
Issue #574 sort order

### LINKED ISSUE #574: BarChart component:  Change the default sort order to be descending by dataValue

jaronheard:
Supercedes #645, resolves #574


jaronheard:
Since this PR targets the component-library, which is being deprecated in favor of smaller packages, I'm closing this PR. A new PR could use this code targeting files in ui-charts

Issue #781, opened 2019-08-15
Add /card/slug routes to mock-wrapper


jaronheard:
### What

Project packages are designed to be run in isolation -- you shouldn't have to run the `2018` package in order to develop in a project package. This is what the `mock-wrapper` package is for.

With the full view of a card that lives at `/cards/slug` (and `/cards/slug/embed`) taking heightened importance, we should incorporate these routes into mock-wrapper so that a developer can preview that functionality.

### Why

So that development can happen within a project package instead of having to run the 2018 package

### When

Not time sensitive, but would be useful as soon as possible

PR #831, opened 2019-08-22
Issue #781

cosmolightfoot:
Moved CardDetailPage and CardDetailPageEmbed from 2018 package to component library. /cards/building-boom-in-portland/ and /cards/building-boom-in-portland/embed both render locally with the new changes.


Dianna:
Adresses #781

packages/2018-neighborhood-development/src/components/ExploreUrbanCampsiteSweeps/index.js
```diff
@@ -64,11 +64,8 @@ export class ExploreUrbanCampsiteSweeps extends React.Component {
             <div>
               <p>{contextualDesc}</p>
               <h2>{timer}</h2>
-              <BaseMap
-                initialLongitude={LONG}
-                initialLatitude={LAT}
+              <BaseMap initialLongitude={LONG} initialLatitude={LAT}>
                 initialZoom={ZOOM}
```
Dianna: `initialZoom` was pulled out of `BaseMap` somehow, so that should be added back in

packages/component-library/src/CardDetail/CardDetailPage.js
```diff
@@ -28,7 +27,16 @@ const CardDetailPage = ({ params }) => {
...
+  params: PropTypes.shape({ slug: PropTypes.string.isRequired }),
+  CardRegistry: PropTypes.shape({
+    entries: PropTypes.arrayOf(
+      PropTypes.shape({
+        component: PropTypes.func.isRequired,
+        project: PropTypes.string.isRequired,
+        slug: PropTypes.string.isRequired
+      })
+    )
+  })
```
Dianna: Since we're doing `.isRequired`, that can be added at the end of the `CardRegistry` prop also.

packages/component-library/src/CardDetail/CardDetailPageEmbed.js
```diff
@@ -21,7 +20,16 @@ const CardDetailPageEmbed = ({ params }) => {
...
+  params: PropTypes.shape({ slug: PropTypes.string.isRequired }),
+  CardRegistry: PropTypes.shape({
+    entries: PropTypes.arrayOf(
+      PropTypes.shape({
+        component: PropTypes.func.isRequired,
+        project: PropTypes.string.isRequired,
+        slug: PropTypes.string.isRequired
+      })
+    )
+  })
```
Dianna: Same, could add `.isRequired` here.

ghost (changes requested):
One breaking change needs to be addressed

ghost (commented):
Excellent. We can now continue this PR :)

packages/mock-wrapper/package.json
```diff
@@ -22,7 +22,10 @@
   },
   "license": "MIT",
   "dependencies": {
+    "@emotion/core": "^10.0.16",
```
Dianna: For safety's sake I've been using the same @emotion/core version across packages and using the 2019-template package as my reference. That one is at ^10.0.10. It shouldn't matter because they both have the `^`. @jaronheard, any preference here?

packages/mock-wrapper/src/components/RootPage.js
```diff
@@ -0,0 +1,16 @@
...
+RootPage.propTypes = {
+  children: PropTypes.node
+};
+
+RootPage.displayName = "Root";
+
+export default connect(
+  () => ({}),
+  () => ({})
+)(RootPage);
```
Dianna: You can remove these two empty functions and just make it:
```javascript
export default connect()(RootPage);
```

packages/mock-wrapper/package.json
jaronheard: I generally do the same thing as @Dianna. It shouldn't matter but might as well keep it consistent.
```suggestion
    "@emotion/core": "^10.0.10",
```

ghost (changes requested):
This is awesome progress and you are super close @cosmolightfoot. This is working very well, but it also needs to support the `2018-<...>` packages (we don't need to worry about any additional packages beyond that though. I think the easiest way to support them would be to make the `CardRegistry` prop and anything else derived from it optional in the `mockWrapper/src/index.js`. My one, significantly less important, comment beside that would be to change the emotion/core version to 10.0.10 and make sure the yarn.lock file doesn't update

packages/mock-wrapper/src/components/RootPage.js
Dianna: Info on [what's going on here](https://stackoverflow.com/questions/41044121/why-call-connect-function-without-mapstatetoprops-or-mapdispatchtoprops-as-args/41044583) if you're curious.

ghost (approved):
Awesome work!

PR #855, opened 2019-08-28
Orb Bounds and Orb Optimize

lucastswick:
The OrbManager is now aware of ORB_CONFIG.verticalBuffer, which ensures `<Orb />` components stay away from the vertical bounds of the container. 

The OrbManager pulls `<Orb />` components closer to the vertical center of the container when they approach the vertical bounds of the container. This prevents a 'hard edge' of the verticalBuffer.

The OrbManager does not slowly push `<Orb />` components downwards.

`<Orb />` components have been refactored to not render on every frame, which was caused by inline callbacks (setOrbTouched, setOrbComplete, addOrbScore) being redefined every frame in the `<OrbManager />`. The refactor is simply to wrap the functions in `useCallback` so the callbacks are properly memoized. The `<Orb />` components are also not passed the entire orb model; if the orb model changes the component would unnecessarily re-render.

### DIFF
--- a/packages/project-disaster-trail/src/components/Game/KitScreen/KitItem.js
+++ b/packages/project-disaster-trail/src/components/Game/KitScreen/KitItem.js
@@ -25,7 +25,7 @@ const KitItem = ({ emptySvg, fullSvg, playerKit, itemType }) => {
     if (playerKit[itemType] && !filledItem) {
       setFilledItem(true);
     }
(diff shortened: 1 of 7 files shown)


ghost (approved):
My dual-core thanks you for these performance optimizations. Also, the "force" orb magic / math is amazing!

PR #858, opened 2019-08-28
Issue 857 - Chart loading skeleton

### LINKED ISSUE #857: Implement MaterialUI skeleton for chart loading
### What

Add MaterialUI Skeleton to ChartContainer for loading state

### Why

UI Improvement

### When

Not time sensitive, but ASAP

jaronheard:
Resolves #857.

This adds a loading skeleton to charts.

Prior to this PR, charts did not have a height before they were loaded, so it caused a page reflow.

| Before  | After |
| ------------- | ------------- |
| Screen Recording 2019-08-28 at 2 25 55 AM  | Screen Recording 2019-08-28 at 2 22 22 AM  |
 
That said, there's a number of things on this PR that I'm not especially confident in.

- This adds a dependency on @material-ui/labs, as the Skeleton component is not yet in the @material-ui/core. Rather than add a dependency, should we just implement this as an individual component?
- Should the Skeleton be its own component? Or is it ok to just have it be implemented in ChartWrapper?
- I added a check for `safeData` to avoid having the charts throw errors if `data` is null and `isLoading` is true, but it smells fishy to me.

### DIFF
--- a/packages/2019-housing/src/components/HomeAppreciation/HomeAppreciationVisualization.js
+++ b/packages/2019-housing/src/components/HomeAppreciation/HomeAppreciationVisualization.js
@@ -91,12 +91,14 @@ const HomeAppreciationVisualization = ({ data }) => {
         xLabel="Home Ownership Rate"
         yLabel="Race"
         dataValueFormatter={x => civicFormat.percentage(x)}
+        protect
       />
       <strong style={{ color: "crimson" }}>
         LineChart Visualization TODO:
         <ul>
           <li>
             Make the confidence interval lines dashed & all lines the same color
+            (see note on lineChartData)
           </li>
         </ul>
       </strong>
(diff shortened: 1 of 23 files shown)


ghost (changes requested):

Dianna:
We should have the `isLoading` prop in all visualizations so we don't have to do the conditional chart mount which causes a layout jump. If that's done here, I'll update the cards in #873 to use it, otherwise if mine is merged first we'll need to do a sweep later and update them.

Dianna:
Height ratio is off. Also, to keep this user friendly we should show the user _something_ besides just a grayish rectangle.

jaronheard:
Current status for this PR:

- It adds a skeleton with the proper height to ChartContainer.
- It resolves #914 

It adds a note about a newly discovered issue with LineChart, which will be covered in another PR:
If there are lines that have different numbers of data points, then they will be colored incorrectly.

ghost (approved):
Loading "skeleton" ratio now works and has a spinner to indicate somethings happening. Tests and documentation look good for protectData

PR #863, opened 2019-08-28
Feature/authors list

Dianna:
Update the author section CivicCard component, metadata, and layouts to handle a list of authors as an array of objects like so:
```javascript
authors: [
    {
      name: "Dustin Henderson",
      email: "scicampwinner07@gmail.com"
    },
    {
      name: "Jim Hopper",
      email: "jim.hopper@yahoo.com"
    },
    {
      name: "Joyce Byers",
      email: "byers0180@hotmail.com"
    }
  ]
```
I've commented out the previous authors code in case it's something we want for reference later, but I can quickly remove it if that's preferred.

Ran storybook, 2019-template, and 2019-transportation packages to check its working.

### DIFF
--- a/packages/2019-template/src/components/DemoCard/demoCardMeta.js
+++ b/packages/2019-template/src/components/DemoCard/demoCardMeta.js
@@ -509,18 +509,9 @@ const demoCardMeta = (/* data */) => ({
     }
   ],
   authors: [
-    {
-      name: "Dustin Henderson",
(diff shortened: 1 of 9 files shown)


jaronheard:
We can remove the names, just take a string of emails. We're using emails as unique ids for lookup.

Jaron Heard
Creative Director
Civic Software Foundation

+1-971-998-7180

Issue #865, opened 2019-08-29
Migrate Disaster 2018 Story Cards to new format


Dianna:
Dianna knows the details ;)

Dianna:
Heh. Dianna doesn't remember the details @jaronheard 🙃

Dianna:
Looks like it was done in #873 though so I'm closing!

PR #873, opened 2019-08-29
Feature/update cards

Dianna:
Addresses #865


Dianna:
Addresses #865
Had to add again to trip github's auto actions...

packages/2018-example-farmers-markets/src/components/PortlandFarmersMarkets/portlandFarmersMarketsMeta.js
```diff
@@ -0,0 +1,135 @@
...
+        { link: "http://www.hackoregon.org", description: "Hack Oregon" },
+        {
+          link: "https://www.civicsoftwarefoundation.org",
+          description: "Civic Software Foundation"
+        },
+        { link: "https://www.civicplatform.org", description: "Civic Platform" }
+      ]
+    }
+  ],
+  // authors likely an array of keys in the future
```
jaronheard: Should this be an empty array now that #874 has been merged?

packages/2018-example-farmers-markets/src/components/FarmersMarketsOverTime/farmersMarketsOverTimeMeta.js
```diff
@@ -0,0 +1,138 @@
...
+const FarmersMarketsOverTimeMeta = (/* data */) => ({
+  title: "Farmers Markets Trending Upward",
+  slug: "farmers-markets-over-time",
+  introText: (
+    <p>
+      Farmers&apos; markets saw steady growth through the 1990s into the
+      mid-2000s. The recession correlates with abnormal growth in the total
+      number of Farmers&apos; Markets. The last two years have shown no growth.
+      Is this plateau expected to continue? What causes growth or decline in
+      Farmers&apos; Markets?
```
jaronheard: One way to prevent having to escape all of the HTML characters and make the text more readable and approachable is like so:
`<p>{/*template string here*/}</p>`

jaronheard: This could be an improvement to our template in making it much easier to copy/paste content. Or we could just have people ESLint fix all fixable problems and end up with a bunch of characters like these

PR #878, opened 2019-08-30
Feature/update api template

Dianna:
### API card generator
Updates the API card generator to correctly update the registry.js rather than the index.js directly. Move the `isLoading` value from <Component>.js to <Component>Visualization.js.

### 2018-disaster-resilience
Remake Proactive Planning card with card generator

### CivicCard
Move `DataChecker` messaging from DOM to console.
Make all cardMeta options (except title) optional _to render_. As in, the card won't break if a section isn't included in cardMeta or is null, but the warning will show in the console.


_templates/api-story-card/with-prompt/src/components/index.ejs.t
```diff
@@ -14,6 +13,7 @@ import api from "../../state/<%=slug%>/api";
 const <%=StoryCardName%> = ({ init, data, Layout }) => {
   useEffect(() => {
     init();
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
jaronheard: >> https://github.com/facebook/create-react-app/issues/6880

packages/2018-disaster-resilience/src/components/App/index.js
```diff
@@ -49,7 +58,9 @@ const App = () => (
     </p>
     {/* <IncreasingSocialCapital /> */}
     <PullQuote quoteText="The #1 thing you can do to increase social capital is to meet your neighbors. Do you know 3 people within a 3 block radius of your house?" />
-    <ProactivePlanning />
+    <section className={sectionMarginMedium}>
```
jaronheard: should this be `css` and not `className`?

packages/2018-disaster-resilience/src/state/proactive-planning/data-parser.js
```diff
@@ -0,0 +1,23 @@
...
+
+const processFeatures = features =>
+  features.map(nbhd => ({
+    displaced_percap: parseFloat(nbhd.displaced_percap),
+    census_response_rate: (100 - parseFloat(nbhd.census_response_rate)) / 100,
+    total_population: parseFloat(nbhd.total_population),
+    quadrant: nbhd.quadrant,
+    resilienceLabel: `${civicFormat.titleCase(
+      nbhd.name
+    )} • Census Non-Response Rate`,
```
jaronheard: This is not an ideal location for something that is a chart label to live. We should look out for this, and fix the underlying issue, which I think is that the chart labels aren't well set up to display category labels

packages/2018-disaster-resilience/src/state/proactive-planning/api.js
```diff
@@ -0,0 +1,26 @@
...
+
+const HOST = "https://service.civicpdx.org/disaster-resilience";
+
+// API DESCRIPTION
+const apiDesc = {
+  getProactivePlanningData: {
+    url: `${HOST}/api/DisasterNeighborhoodView/?limit=150`,
+    // you can apply any needed data transformations to value here
+    // if complex, separate tranformation function to another file
+    dataTransform: data => getProactivePlanningData(data.results)
```
jaronheard: Data transformation, nice!

_templates/api-story-card/with-prompt/src/components/index.ejs.t
Dianna: I think it would take me awhile to figure out what to do in this case and isn't the best use of my time currently. But there is the chance that I don't understand the danger well enough, so if you think that's the case let me know and I will research.

Desired effect: `init()` is run when the component mounts (like `componentDidMount` in older react). The `[]` second argument is used to prevent the function from being called more than once. Once the data has been fetched, it never needs to be fetched again.

packages/2018-disaster-resilience/src/components/App/index.js
Dianna: 🙀YES

packages/2018-disaster-resilience/src/state/proactive-planning/data-parser.js
Dianna: I'd love to learn more about this. In this case I was just copying over the data transformations as is so I didn't unknowingly break something.

jaronheard: I'm filing an issue for this.

PR #895, opened 2019-09-01
Feature/authors demo

Dianna:
Make CivicCardLayoutFull handle 3 values for authors: undefined / null / empty array, 'demo', and an array of author emails

### DIFF
--- a/packages/component-library/src/CivicCard/cardMetaTypes.js
+++ b/packages/component-library/src/CivicCard/cardMetaTypes.js
@@ -1,7 +1,7 @@
 import PropTypes from "prop-types";
 
 const cardMetaObjectProperties = {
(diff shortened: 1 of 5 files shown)


packages/component-library/src/CivicCard/cardMetaTypes.js
```diff
@@ -28,7 +28,10 @@ const cardMetaObjectProperties = {
     })
   ),
   // authors: PropTypes.arrayOf(PropTypes.string /* image url */)
```
jaronheard: can delete this comment now

Dianna: Woops. Meant to. It's covered in the condition below

PR #900, opened 2019-09-03
Design color update

jaronheard:
Updates design colors

Sets tertiary to Plum (primary)
Formalizes action color


packages/component-library/src/_Themes/Brand/BrandColors.js
```diff
@@ -11,11 +11,11 @@ export default {
     rgba: "rgba(238,73,92,1)",
     hex: "#EE495C"
   },
-  // Plum Light
+  // Plum
```
Dianna: This should be renamed...

Dianna: per the other "plum" above

Dianna: Oh right. NVM, supposed to be the same

PR #903, opened 2019-09-03
A new button component for the Civic Platform

bradfordjanice:
Created a new button component from the Material UI button component.  This was done so it will be consistent in UI behavior and style to CIVIC's other input components.

It has been added to Storybook.

CIVIC visual styles will be updated with all the other input components in a separate pull request.


packages/component-library/src/ButtonNew/ButtonNew.test.js
```diff
@@ -0,0 +1,41 @@
+import React from "react";
```
jaronheard: You can remove this test file for now, these failing tests are causing Travis to fail

Issue #728, opened 2019-08-05
Enforce consistent Yarn version


jaronheard:
### What
**[Enforce Yarn’s version across your project](https://yarnpkg.com/en/docs/cli/policies)**

### Why

Inconsistent yarn versions can lead to dependency issues

### When

ASAP

PR #908, opened 2019-09-05
Chore/restart housing 2019

BJHunnicutt:
🚧This is a WIP PR for Housing Cards so I can get feedback from the rest of the team
- Rebuild the 2019-housing package based on the up-to-date template generator
- Build out ALL Housing Story cards as far as I can (still working on Card #9)
- Add notes to remaining tasks/decisions for all visualizations


yarn.lock
```diff
@@ -1174,7 +1174,6 @@
 "@emotion/cache@^10.0.15":
   version "10.0.15"
   resolved "https://registry.yarnpkg.com/@emotion/cache/-/cache-10.0.15.tgz#b81767b48015aae2689c60373992145c67b8de02"
-  integrity sha512-8VthgeKhlGeTXSW1JN7I14AnAaiFPbOrqNqg3dPoGCZ3bnMjkrmRU0zrx0BtBw9esBaPaQgDB9y0tVgAGT2Mrg==
```
jaronheard: @BJHunnicutt can you upgrade your version of yarn - the integrity checks are being stripped from the yarn.lock file.

Once we resolve #728, we shouldn't have to worry about this anymore. We can check in a version of yarn using yarn-path as described here to avoid this issue in the future. More info: yarnpkg/yarn#6440

packages/2019-housing/src/components/HouseholdIncomeByRace/HouseholdIncomeByRaceVisualization.js
```diff
@@ -0,0 +1,55 @@
...
+    { category: "White", label: "White" }
+  ];
+
+  return (
+    data && (
+      <span>
+        <strong style={{ color: "crimson" }}>
+          Visualization TODO:
+          <ul>
+            <li>Figure out why dot colors do not match lines...</li>
```
jaronheard:
```
// VictoryPie uses d3-interpolate to perform animations. d3-interpolate
// “sanitizes” all values, despite whether or not they are used for an animation.
// This results in english words that happen to also be valid html/css color names
// become rgb(x, x, x) strings.
const protectLabels = data => data.map(d => ({ ...d, name: `${d.name}.` }));
```

jaronheard: This might be helpful for why the colors don't match lines

PR #910, opened 2019-09-05
Material UI Theme for Radio Button Group

bradfordjanice:
Set up the file structure for the Material UI Theme which will be used in all the Material UI components.
First application is to the Radio Button Group.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,9 +66,7 @@
       "git add"
     ]
   },
-  "dependencies": {
-    "@material-ui/styles": "^4.3.3"
-  },
+  "dependencies": {},
   "devDependencies": {
     "@babel/cli": "^7.0.0",
     "@babel/core": "^7.0.0",
--- a/package.json
+++ b/package.json
@@ -66,7 +66,9 @@
       "git add"
     ]
   },
-  "dependencies": {},
+  "dependencies": {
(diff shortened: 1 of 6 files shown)


packages/component-library/src/_Themes/index.js
```diff
@@ -1,9 +1,7 @@
 export { default as VisualizationColors } from "./VisualizationColors";
-
 export { default as BrandColors } from "./Brand/BrandColors";
-
 export { default as BrandTheme } from "./Brand/BrandTheme";
 export { default as VictoryTheme } from "./Victory/VictoryTheme";
 export { default as VictoryCrazyTheme } from "./Victory/VictoryCrazyTheme";
-
+export { default as muiTheme } from "./MaterialUI/MaterialUITheme";
```
jaronheard: Can we rename this MaterialTheme for consistency with our other themes?

package.json
```diff
@@ -66,7 +66,9 @@
       "git add"
     ]
   },
-  "dependencies": {},
+  "dependencies": {
+    "@material-ui/styles": "^4.3.3"
```
jaronheard: Looks like this got added at the workspace root, it should just be a dependency of the `component-library`. From the workspace root, can you `yarn remove @material-ui/styles`

PR #911, opened 2019-09-05
Remove --fix from lint hook

jaronheard:
This changes functionality of the commit hook to spit back out fixable eslint errors instead of fixing them automatically.

Fixing the errors automatically could result in unexpected behavior, which I witnessed @worc have an issue with


jaronheard:
Decided that changing this functionality between now and Demo Day might be too disruptive. Note to revisit after Demo Day

PR #912, opened 2019-09-05
Disturbance stops comparison map, proof of concept

worc:
This works, usually. It's not quite ready for a real deploy; it's using axios directly to do paged API calls so that TOAD doesn't fall over.

Next steps for this card is to parameterize a reduxful config and add UI knobs for those parameters


jaronheard:
This doesn't break anything, and will give people stuff to review, so ... ⛴📦⛴

PR #913, opened 2019-09-06
Export MaterialUI Theme from component-library

jaronheard:
So that it can be used in other packages, similar to other themes.


bradfordjanice:
Looks good!

PR #918, opened 2019-09-06
Summary screen WIP

Dianna:
I'm not quite sure how to animate this properly. It would be cool if the top text animated in quickly with the secondary text on its tail and if they left the screen the same way. After these all finish, the game should return to the attractor screen


Dianna:
superseded

PR #925, opened 2019-09-07
Improve slider & add titles to ComparisonMap component

BJHunnicutt:
- Update the ComparisonMap slider so it is more clear that the map is interactive. It used to be a slider below the map & now it is an anchor with arrows on it between the two maps.
Also added:
- Optional Title components for each map, which get covered if the map slider passes them (leftMapTitle & rightMapTitle props)
- Optional divider between the two maps to make a clear distinction if the color scales are the same on both sides (showDivider prop)

New ComparisonMap:
Screen Shot 2019-09-07 at 1 16 09 PM
Screen Shot 2019-09-07 at 1 16 25 PM

Old Version:
Screen Shot 2019-09-07 at 1 18 12 PM

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1054,7 +1054,6 @@
 "@babel/runtime@^7.5.5":
   version "7.5.5"
   resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.5.5.tgz#74fba56d35efbeca444091c7850ccd494fd2f132"
-  integrity sha512-28QvEGyQyNkB0/m2B4FU7IEZGK2NUrcMtT6BZEFALTguLk+AUT6ofsHtPk5QyjAdUkpMJ+/Em+quwz4HOt30AQ==
   dependencies:
     regenerator-runtime "^0.13.2"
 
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -1054,7 +1054,6 @@
 "@babel/runtime@^7.5.5":
   version "7.5.5"
   resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.5.5.tgz#74fba56d35efbeca444091c7850ccd494fd2f132"
-  integrity sha512-28QvEGyQyNkB0/m2B4FU7IEZGK2NUrcMtT6BZEFALTguLk+AUT6ofsHtPk5QyjAdUkpMJ+/Em+quwz4HOt30AQ==
```
BJHunnicutt: Not sure why this happened again... fixing it now.

BJHunnicutt:
@mendozaline, do I need to update the version of the component-library to get this update to take effect?

jaronheard:
@BJHunnicutt You should be able to build the component-library to get the changes

BJHunnicutt:
@jaronheard, Totally fine changing them, but wanted to point out the reasons I initially put them on the map. 
1. To reinforce which one they were associated with even if one map is slid over to the point that it is under the title (when they're on the map, they get covered by the slider)
2. So they didn't interfere with the ChartContainer content

Here are comparisons for those 2 things:

Screen Shot 2019-09-07 at 3 39 46 PM
vs
Screen Shot 2019-09-07 at 3 41 32 PM

jaronheard:
Great considerations @BJHunnicutt !

I was concerned with the legibility of the text when they're inside of the map, especially for dark mode.

Another option would be a background box for the text in the map.

I fiddled with a few ideas that were kind of combination but down for anything!
Screen Shot 2019-09-07 at 4 28 33 PM
Screen Shot 2019-09-07 at 4 27 40 PM
Screen Shot 2019-09-07 at 4 27 14 PM

This is super rad!

BJHunnicutt:
I'm going to merge it with the labels above the maps, and we can follow up with your additions if it ends up being confusing

PR #930, opened 2019-09-08
Add Education Sample Card

jaronheard:
Card uses demo data for education data collaborative site
There's a theme added that switches the order of the data visualization colors for the education data collaborative site


packages/2019-education/src/common/DataCollaborativeVictoryTheme.js
```diff
@@ -0,0 +1,313 @@
+import { assign } from "lodash";
```
Dianna: This should be imported from the component library rather than recreated unless it's style is different (which I can't tell off-hand) or we want for this package to avoid brand / visualization updates that the rest of the site receives.

packages/2019-education/src/common/DataCollaborativeVisualizationColors.js
```diff
@@ -0,0 +1,108 @@
+import BrandColors from "./DataCollaborativeBrandColors";
```
Dianna: These should be imported from the component library rather than recreated unless we want for this package to avoid brand / visualization updates that the rest of the site receives

jaronheard:
Cleaned up the theme -- there wasn't a great reason to fully recreate the theme for a few overrides

Issue #10, opened 2017-10-22
Fix broken components in budget project


clemf:
Budget builds but many components and styles are broken

DavideDaniel:
This is partly due to changes that were made for emergency response to the core components but the other half of the styling might be in emergency response itself. We should make sure they are together.

PR #933, opened 2019-09-08
Chore/clean up housing cards

BJHunnicutt:
More updates to Housing 2019 Cards. Still a 🚧work in progress 🚧, but there are a lot of good intermediate changes I wanted to get feedback on before I continue.


Dianna:
Card #2: Racial Demographics of Home Loan Approvals
- Home loan approval rates needs a legend (what is green vs brown?)
- Maybe set a max zoom in / out?

Card #4: Black Population Share Change Over Time
- Slider button could be more obvious
- I think a 1 line explanation would be good (i.e., Slide to compare 1990 to 2017)
- Maybe set a max zoom in / out?

Card #6:
- Maybe set a max zoom in / out?
- Top 2 maps are the same data right, just different implementations? I like the top one, with tooltips
- Bottom visualization is scatterplot without enough data or line graph without a line?

Card #8:
- Make sure to add legend when this map is available

Card #10: Households of Color Earn Substantially Less than White Households
- Dot colors match lines now. Was a chart implementation issue

packages/2019-housing/src/components/BlackPopulationChange/BlackPopulationChangeVisualization.js
```diff
@@ -5,13 +5,18 @@ import { scaleQuantize, extent } from "d3";
 
 import {
```
Dianna: In `packages/2019-housing/src/components/AduDistributions/index.js`, the eslint-disable is not working for init, needs to be repositioned

Dianna:
In `AduDistributions/index.js`, `HolcRedlining/index.js`, `HomeAppreciation/index.js`, `HomeLoanApprovals/index.js`, `HomeOwnershipRates/index.js`, `HouseholdIncomeByRace/index.js`, and `HousingDisplacement/index.js` the eslint-disable is not working for init, needs to be repositioned.

Or this works:

```
useEffect(() => {
    init();
  }, [
    /*
    https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects

    Add second argument to prevent useEffect running init() again
    */
  ]);
```

BJHunnicutt:
@Dianna, I think I addressed most of your comments, I updated the init() thing everywhere and added min/max zoom to all the maps.

Card 2: My next task is to build a MapLegend (which will affect most of the cards), but I want to get these changes in before I do that. 

Card 4: I am planning to work a bit more on the compare map to make the slider more obvious, but in the meantime I added what you recommended to the subtitle of the map.

Card 6: Housing had asked me to make both maps & they also picked the top 1. The scatter plot was just a stand-in/fake data because the API wasn't ready yet, but I removed it for now.

(**Side note**: If you test this again, Card 2 will infinitely load because they are changing the API)

jaronheard (approved):
If you want to merge this so that others can see to review, go ahead when you please :)

PR #936, opened 2019-09-08
2019 transportation/disturbance line chart

worc:


packages/2019-transportation/src/components/DisturbanceStops/disturbanceStopsLineChartMeta.js
```diff
@@ -1,20 +1,19 @@
...
 
-import TransportationCardVisualization from "./TransportationCardVisualization";
+import DisturbanceStopsLineChartVisualization from "./DisturbanceStopsLineChartVisualization";
 
-const transportationCardMeta = (/* data */) => ({
-  title: "(Story Card #1, Issue #000) Transportation Card Title",
-  slug: "transportation-card",
+const DisturbanceStopsMeta = (/* data */) => ({
+  title: "Total Time of Disturbance Stops",
+  slug: "disturbance-stops",
```
Dianna: This should have a different slug than the other card. You can see if you click "Share" and go that link, that it will direct you to the non-line-chart card

packages/2019-transportation/package.json
```diff
@@ -44,6 +44,7 @@
     "d3-array": "^2.3.1",
     "express": "^4.14.1",
     "invariant": "^2.2.2",
+    "moment": "^2.24.0",
```
Dianna: Use [date-fns](https://date-fns.org/) instead. It will be a super easy switch. Make sure to use `yarn remove` to get rid of `moment` so that it cleans up the `yarn.lock` file

packages/2019-transportation/src/components/DisturbanceStops/DisturbanceStopsVisualization.js
```diff
@@ -1,10 +1,10 @@
-import React, { useEffect, useState } from "react";
-import axios from "axios";
+import React, { useEffect, useState, useContext } from "react";
+import moment from "moment";
```
Dianna: use [date-fns](https://date-fns.org/)

packages/2019-transportation/src/index.js
```diff
@@ -3,17 +3,12 @@ import Routes from "./routes";
...
 
 const CardRegistry = [
   {
     slug: "disturbance-stops",
     component: DisturbanceStops
-  },
-  {
-    slug: "transportation-card",
-    component: TransportationCard
   }
```
Dianna: Need to add a slug for the other card if you decide not to combine them

packages/2019-transportation/src/components/DisturbanceStops/index.js
```diff
@@ -1,46 +1,82 @@
-import React, { useEffect } from "react";
+import React, { useState, useEffect } from "react";
+import axios from "axios";
```
Dianna: All data requests should be done in the state directory. If you run `yarn card` from the root of the repo, it will build the whole card structure for you so all you have to do is add your URL and sub in your visualization and card info. You should be able to copy paste your stuff from these cards

ghost (changes requested):
Ohhhhkay. So I talked with Jaron and understand that your app has different needs than most, so I'll close out most of my comments. What we must do though is find a solution for share. Currently, the share button on either card only opens one of the civic cards which is weird. If these two cards are bound, then they should embed together and you will need to create a custom component for that. We can chat about it over a call if you like to figure out what will work best.

packages/2019-transportation/src/components/DisturbanceStops/index.js
Dianna: Nevermind. I understand you have a special use case

worc: yeah i started with axios/hooks to do a quick and dirty paging/params solution. the API endpoints seem to have a little more power behind them now, so i think it should be good to migrate from axios/hooks back into reduxful

packages/2019-transportation/package.json
worc: cool working on this now. date-fns doesn't have support for parsing durations, as in going from a string `mm:ss` to an integer or decimal value of seconds or minutes:

https://github.com/date-fns/date-fns/issues/284

so, now that i think about it, it feels more like something that should be handled on the API side. our views really want to handle numbers more than duration strings.

for now i think i'll write a quick duration parser and then also see what it would take to update the API.

packages/2019-transportation/src/components/DisturbanceStops/disturbanceStopsLineChartMeta.js
worc: for this, i brought both visualizations together into the same card. now there's only one card and share link between the comparison map and the line chart

jaronheard (approved):
Looks good! At some point would like to migrate these to reduxful, if that's not blocked, but not the highest priority.

ghost (approved):

worc:
@jaronheard yeah definitely want to put migrating back into reduxful on the todo list. the piece that i wasn't able to get around to this summer was looking into how reduxful would handle automatically paging. it's not strictly a front-end concern to not topple the API, but it'd be good to be responsible with the way we chunk in the data and take advantage of that `next` field in the responses

PR #937, opened 2019-09-08
Feature/fix orb completion

lucastswick:
Fixes `<Orb />` completion events
Makes `<Orbs />` charge in 2 seconds.


packages/project-disaster-trail/src/components/atoms/Orb.js
```diff
@@ -102,7 +102,7 @@ export default class Orb extends PureComponent {
         this.setState({ isComplete: true });
         clearInterval(timer);
       } else if (percent < 100 && isActive) {
-        const update = percent + 10;
+        const update = percent + 20;
```
Dianna: I'm actually likely going to remove this whole function. RadialGauge doesn't need it

PR #940, opened 2019-09-09
Feature/sounds

lucastswick:
Added a `<Song />` component that plays an audio file when mounted, and fades out when unmounted. All chapters have an audio track, but note that the Tasks Chapter does not change audio files when switching from Solving to Voting. 

Added a `SFX` reducer with a `playSFX()` action that can be triggered from any component to play a sound effect. Current SFX are the 'tick' on the start button, and a good orb sound (note that all orbs play a good sound at the moment)


packages/project-disaster-trail/src/state/sfx.js
```diff
@@ -0,0 +1,27 @@
+import { createReducer } from "redux-starter-kit";
+// import { Howl, Howler } from "howler";
```
Dianna: Can remove all of this commented out code in this file

packages/project-disaster-trail/src/components/Game/index.js
```diff
@@ -16,6 +16,7 @@ import KitScreen from "./KitScreen/index";
 import TaskScreen from "./TaskScreen/index";
 
 import "@hackoregon/component-library/assets/global.styles.css";
+import quakeSong from "../../../assets/audio/PWolfAfter_Earthquake1fadeinout.mp3";
```
Dianna: This song is too dark I think

packages/project-disaster-trail/src/components/Game/TaskScreen/index.js
```diff
@@ -19,6 +19,8 @@ import Timer from "../../../utils/timer";
 
 import * as ACTIONS from "../../../constants/actions";
 import usePrevious from "../../../state/hooks/usePrevious";
+import taskSong from "../../../../assets/audio/PWolfAfter_Earthquake1fadeinout.mp3";
```
Dianna: This song is too dark. We should use a different one

PR #954, opened 2019-09-10
Sandbox 2.0

mendozaline:
This PR contains the updates to the Sandbox:
- Updates the backend api endpoint
- Makes the map loads first, not the package selector screen
- Lint most of the `civic-sandbox` package and components
- Implements fit bounds to the Base Map
- Memoizes centroid calculation for bubble maps


mendozaline:
Fixed the focus outline issues on the side panel. 

Not sure about the data issue. I was able to confirm that the lowest value for that key was 397:
Screenshot_2019-09-09

Issue #877, opened 2019-08-30
Separate out SandboxLegend into reusable component


jaronheard:
### What

SandboxLegend is tied up into the Sandbox implementation.

We should create a new MapLegend component that is functionally identical, but can be used for both the Sandbox and not Sandbox.

### Why

So that we can show a map legend with MultiLayerMap in CivicCard

### When

ASAP, before Demo Day

jaronheard:
#962 addressed

PR #962, opened 2019-09-10
Add a MapLegend component

BJHunnicutt:
This is a first pass at creating a map legend component & pulling it into the Housing Story Cards.

It definitely needs to be fleshed out in the future for more use cases, but so far it handles:
- Only requires passing in the same `colorScale` function used for the maps
- Categorical or ranged data
- Vertical or horizontal orientations
- Has a label option
- Takes a function to format the tick values

Here are a few examples:
Screen Shot 2019-09-10 at 2 48 05 AM
Screen Shot 2019-09-10 at 2 48 16 AM
Screen Shot 2019-09-10 at 2 47 47 AM
Screen Shot 2019-09-10 at 2 47 36 AM


BJHunnicutt:
Issue #877

PR #967, opened 2019-09-10
Update Education Page

jaronheard:
Cleans up Education page to share with Education team

### DIFF
--- a/packages/2019-education/src/components/App/index.js
+++ b/packages/2019-education/src/components/App/index.js
@@ -6,7 +6,7 @@ import {
   BrandTheme,
   PageLayout,
   PullQuote,
-  CivicCardLayoutClassic // eslint-disable-line
+  CivicCardLayoutVisualizationOnly // eslint-disable-line
 } from "@hackoregon/component-library";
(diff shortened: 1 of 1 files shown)


bradfordjanice:
The page looks great!  Nice summary and presentation of the data!

I found a couple of things:
Problem with some buttons on the Learn More page.  To replicate:
* Click on `Learn more` link below the Navigator Program Participant Achievement Growth, this takes you to a page with the header `Is There Evidence the Program Worked?`
* Scroll down to the `Make this better` section and hover over the two buttons:  `GET INVOLVED!` and `FILE AN ISSUE`.
* On hover, the button labels disappear.
* The buttons are still active.
* The similar buttons on the other pages reached by clicking `Learn more` have similar behavior.

I don't know if this is because I'm working in the development environment instead of production environment, but when I click the browser back button, as in the buttons described in my first comment, the previous page doesn't reload. The URL does change back to the previous page, but it is blank; refreshing the browser still gives a blank page.  I'm seeing this throughout... clicking the back button isn't working.

PR #968, opened 2019-09-10
add the year as the comparison map titles

worc:
i had to throw an important tag on these. CivicStoryCard's h2 style is more specific than the style defined in the comparison map, overriding it.


BJHunnicutt:
I just realized that there's an issue with the `height` prop in the ComparisonMap -- if you add `height={500}` to all of these ComparisonMaps, you'll get rid of that weird tail. I'll look into fixing the issue after demo day

worc:
cool, added those two things. maps don't have the slider tail anymore and have a little bit of breathing room afterwards for the next element

PR #982, opened 2019-09-11
2019 transportation/follow the demo outline

worc:
getting the cards _literally_ in order


worc:
this is getting a little large so i'm going to stop here. i removed the dummy text so that worst-case-scenario we demo something without the lorem-ipsum-ice-cream in the individual card views. and there's also now a shared additional text for the comparison maps.

final final final text and edits will come from here:
https://docs.google.com/document/d/1J5BgOUgJoB5TDL6lNX5xAKJiFvJ2q7P413b4XmIoEMU/edit#

and then will have to be committed/merged as need be.

packages/2019-transportation/src/components/SystemWideSummary/SystemWideSummaryVisualization.js
```diff
@@ -0,0 +1,157 @@
...
+            xLabel="Quarter Hour"
+            xNumberFormatter={time => {
+              return time
+                ? `${Math.floor(time)}:${
+                    (time % 1) * 60 === 0 ? "00" : (time % 1) * 60
+                  }`
+                : time;
+            }}
+            yLabel="Ons and Offs"
+          />
```
jaronheard: Can you add a comment to say why this is commented out?

worc: yes definitely. i had half a thought to add another stacked line chart with the delay time buckets: fifth percentile, quartile, median etc. but i'm not sure it necessarily follows the demo outline that the team is writing.

i'll add that as a note.

PR #1006, opened 2019-09-11
Chore/housing content changes

BJHunnicutt:
This is not the end of it, but here are is the MVP for Housing page content...


packages/2019-housing/src/components/AduDistributions/aduDistributionsMeta.js
```diff
@@ -6,7 +6,7 @@ import AduDistributionsVisualization from "./AduDistributionsVisualization";
 const AduDistributionsMeta = (/* data */) => ({
   title: "Card #6: Adu Distributions (TODO: Make a real title)",
   slug: "adu-distributions",
-  introText: <p>INTRO TEXT: TBD</p>,
+  introText: <p />,
```
Dianna: future: could put `null`

BJHunnicutt: done

PR #1043, opened 2019-09-19
make video controls appear on hover or focus

fstorr:


Dianna:
I was running into an issue with being able to show a poster for this PR on certain browsers. An easy solution would be to who the white screen at the end of the video rather than the beginning, so we don't even have to implement the poster.

Issue #1037, opened 2019-09-18
State refactor to fix timing a state bugs


Dianna:

Dianna:
Resolved by #1055

PR #1055, opened 2019-09-24
WIP - Refactor the disaster task screen game state

ghost:
There were some fundamental bugs in how the task screen was timing phases of the task screen and using / updating properties. I tried a refactor, but I'm broadly remaking those aspects of the task chapter instead which is much simpler than untangling logic that isn't following best practices anyway.

I threw in some small fixes to remove annoying console errors, and added the ticker tape content since I already had that ready to go.


Dianna:
Resolves #1055

Dianna:
Resolves #1037

packages/2019-disaster-game/src/components/Game/TaskScreen/mapUtils.js
```diff
@@ -74,7 +74,7 @@ export const getPosition = f =>
 export const asGeoJSON = (tasks, activeTask, completedTasks) =>
```
jaronheard: maybe rename this function as it seems to be doing a lot more than asGeoJSON

packages/2019-disaster-game/src/components/Game/TaskScreen/SolveScreen.js
```diff
@@ -1,93 +1,73 @@
...
+      setTaskImageCss(taskImageGenerator(activeTask));
+    }
+  }, [activeTaskIndex, activeTask]);
 
   return (
     <div
       css={css`
         ${screenLayout}
-        ${open ? onScreenStyle : {}}
+        ${open && onScreenStyle}
```
jaronheard: If open is false, this will be interpolated to "false", which could result in invalid css. 

I think you may want `${open ? onScreenStyle : ""}`

jaronheard (approved):
Cool! I tested this and it seems to be is working well. I made a couple of comments. I ignored the extraneous comments per request.

jaronheard:
One other note, which you may be aware of:
image

Dianna:
> One other note, which you may be aware of:
> image

Yup... I'm aware of it. It's due to how orbManager animates things and that's a refactor I'm avoiding if possible...

packages/2019-disaster-game/src/components/Game/TaskScreen/SolveScreen.js
Dianna: Good point. Fixed

packages/2019-disaster-game/src/components/Game/TaskScreen/mapUtils.js
Dianna: Since I didn't need to implement this function and would need to dig in to understand it, I'm going to pass on this for now in lieu of more important tasks

Issue #1026, opened 2019-09-17
Launch game on secret route for OMSI by 9/19


Dianna:
Due date: Thursday morning 9/19
OMSI needs this to stay in step on our progress

Should be in the secret 2019 route within the `2018` package.

Dianna:
Has been completed by some PR or other...

PR #1056, opened 2019-09-25
Fix production build of the disaster game

jaronheard:
Something was going on with the production build of the disaster game, and it was weird. But, this fixes it.

Removing @babel/plugin-transform-react-constant-elements also would resolve this, at a significant hit to performance.

I would *love* to know what's happening here, but it took me forever to track down this issue, and I should move on to other thing.


Dianna:
Resolves #1026

PR #1062, opened 2019-09-25
Change PageLayout component to use main tag

### LINKED ISSUE #779: All pages are missing a `main` section
## What
At the recent hack day, we added HTML to mark up the header and footer landmark areas of each page. The last major area we need to add is the "main" one. The main area must contain the main content of the page, i.e., not the header, footer, navigation bar, or any sidebars. On the [demo card page](https://civicplatform.org/cards/demo-card), this is the `<div class="css-o92fdl">` element.

Adding this element will provide the correct semantics to assistive technology and allow screen reader users to quickly move to the content. We can achieve a main element in one of two ways:

1. Using the HTML `<main>` element.
2. Adding the `role="main" attribute to the existing `<div>`.

The first option is better as it uses native HTML and meets the [First Rule Of ARIA Use](https://www.w3.org/TR/using-aria/#firstrule), which is to use HTML if there is a suitable element available.

While adding this, can we also add `tabindex="-1" and `id="main"` attributes to the element, and then the necessary code to hide the focus ring. We'll need these when we add a "skip link" (issue for this is forthcoming).

### HTML required
Either: `<main id="main" tabindex="-1">` or `<div class="css-o92fdl" id="main" role="main" tabindex="-1">`

### CSS required
`main:focus{outline:none}` (or different target element if using the `<div>`)

## When
It would be good to have this as soon as possible. If possible.

mikaelacurrier:
Hello there :wave:

I wrote this to fix #779. Let me know if you need anything else changed on it. :bowing_woman: 

**Questions**:

- [ ] I ran the build command and the dist/es are ignored, would you still like those included or is that documentation out of date?

### DIFF
--- a/packages/component-library/src/PageLayout/PageLayout.js
+++ b/packages/component-library/src/PageLayout/PageLayout.js
@@ -9,6 +9,10 @@ const headerHeight = 72;
 
 const defaultStyles = css`
   padding: 0px 24px;
+
+  main:focus {
+    outline: none;
+  }
   @media (max-width: 640px) {
     padding: 0px 6px;
   }
@@ -54,7 +58,7 @@ const PageLayout = ({
   teamTitle,
   children
 }) => (
-  <div css={header && headerMargin}>
+  <main id="main" tabIndex="-1" css={header && headerMargin}>
     {header && <Header />}
     {heroTitle && (
       <CollectionHero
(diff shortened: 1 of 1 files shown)


jaronheard:
@mikaelacurrier Look great! Thanks you so much for the PR!

> I ran the build command and the dist/es are ignored, would you still like those included or is that documentation out of date? 

We must have some out of date documentation, where did you see that referenced?

mikaelacurrier:
@jaronheard 

In the contributing.md under Rebuilding The Component Library it states:

```Build with yarn build while in the packages/component-library directory. This will generate changes in the es and dist directories. Commit those changes.```

Issue #1069, opened 2019-09-27
Update "card not found" page


jaronheard:
### What

Update the card not found message seen at http://civicplatform.org/cards/any-non-existent-slug

- [ ] Change link from Portland collections to /cards, with link title "other cards"
- [ ] Put "Card not found" text in section wrapper

Should be updated in these two files:
<img width="1678" alt="image" src="https://user-images.githubusercontent.com/7065695/65748894-c62e3780-e0b9-11e9-89f5-5617ac69393e.png">

### Why

Otherwise it links to old projects

### When

Not time sensitive

amoury:
Hey @jaronheard , can you please provide some clarification for the todo 2 - `Put "Card not found" text in section wrapper`.

Does it need to go inside a div with the className = `section-wrapper`

jaronheard:
#1072 resolved

Thanks @daviccardoso for the PR, and thank you to @amoury for asking about this one as well!

PR #1072, opened 2019-10-02
Update 'card not found' page

daviccardoso:
Update message according to instructions provided in issue https://github.com/hackoregon/civic/issues/1069.


jaronheard:
Awesome! Thanks so much @daviccardoso

PR #1075, opened 2019-10-03
add icons and update sprite sheet

jaronheard:


jaronheard:
@Dianna what’s the source of truth for the icons?

Dianna:
The task icons within the assets folder. It looks like the sprite sheet generator mangled them

Issue #872, opened 2019-08-29
Links in the footer of Civic cards have non-hidden decorative icons


fstorr:
## What

The links in the footer of cards have icons that aren't hidden from assistive technology, which means they could get announced along with the text in the links. Here's a screen shot of VoiceOver's Web Rotor showing two links with question marks in them. Those question marks are VoiceOver's way of saying "there's something here, but I don't know what it is".

![screen shot of VoiceOver's web rotor showing the links panel](https://user-images.githubusercontent.com/169471/63910014-e47d1680-c9d9-11e9-9bc9-443b2b8eb2d2.png)

As the icons are purely decorative (they're not the only content in the link, they're just there to make it look more attractive), we should hide them so that the link only contains the text.

## How
Use `aria-hidden="true" on the `<i>`:

```
<a href="…">
  <i aria-hidden="true" class="fa fa-info-circle"></i><span>Link Text</span>
</a>
```

PR #1080, opened 2019-10-05
add aria-hidden to icons

paigewilliams:
Resolves issue #872 

I added `aria-hidden="true"` to all of the icons in the `i` elements in the `StoryLink` components.


jaronheard:
Thanks @paigewilliams !

Issue #871, opened 2019-08-29
Civic card Share button needs to either be a button or support spacebar activation


fstorr:
## what

The [Civic card](https://hackoregon.github.io/civic/?path=/story/component-lib-story-cards-civic-card--layout-classic) Share "button" has this markup:

```
<a tabindex="0" role="button"><i class="fa fa-link"></i><span>Share</span></a>
```

The ARIA `role="button"` attribute tells assistive technology (e.g. a screen reader) that the element is a button. What it doesn't do is to actually make it behave like a button and work when the space bar is pressed (links are keyboard activated with the Return key, and buttons by the Return key or the space bar).

## How to fix:

Do one of these:

1. Keep the existing code but add JavaScript support for the space bar so runs the function in the same way the Return key does.
2. Replace the existing code with a `<button>` element, which is the correct one for this component:

```
<button type="button">
  <i aria-hidden="true" class="fa fa-link"></i><span>Share</span>
</button>
```

Also note that the icon has the `aria-hidden="true"` attribute, which hides it from assistive technology. This should be added to the link option if that option is used instead of the button.

jaronheard:
#1084 addressed

Issue #823, opened 2019-08-22
Add focus outline to global style


jaronheard:
We need to add focus outlines to our global styles. However, first we need a design for what we want them to look like.

PR #1082, opened 2019-10-05
Share button a11y

paigewilliams:
Addresses issue #871 

In the issue, @fstorr suggests two ways of going about the solution. I decided that changing the element to a button was the best option. I updated the css to match the previous styles.


packages/component-library/src/CivicStoryCard/CivicStoryLink.js
```diff
@@ -4,14 +4,21 @@ import { Link } from "react-router";
...
     align-items: center;
     justify-content: space-between;
     padding: 8px 12px;
     border-bottom: none;
     color: #000;
+    border: none;
+    background-color: #eee;
+
+    &:focus {
+      outline: 0;
```
jaronheard: If we're going to set outline to 0 here, we should have some alternative styles for focus -- read more here for why: http://outlinenone.com/

#823 is a related open issue but needs a design. For now we could leave the focus as default.

Issue #705, opened 2019-08-02
Match-lock: multiple good orbs


Dianna:
ONLY APPLIES TO ONE _TYPE_ OF GOOD ITEM AT A TIME
Example: 2 water orbs chosen will trigger the effect, if 2 food are chosen before water completes its selection cycle, the food items will NOT gain the effect

- [ ] When second of good item type is chosen all items of that type grow 1.5x in size (size is just POC, make sure the gauge expands with the orb)
- [ ] It and any other items of the type that are selected will have their radial gauge start at the same place as the initially touched orb
- [x] On completed selection all the items that were selected fly to the top of the screen

Dianna:
Not doing match-lock

Issue #1078, opened 2019-10-04
Kit screen match-lock item adding


Dianna:
* Match-lock only happens for water, food, and first-aid
* When a normal item is selected, a gauge fills around the upper screen representation
* When water, food, and first-aid are selected singly, or in twos the gauge fills halfway, but cannot be completed in this fashion
* When water, food, and first-aid are match-locked, the gauge will fill the rest of the way

Dianna:
Superseded

PR #1083, opened 2019-10-07
Feature/match lock

Dianna:
Introduces match-locking. It is triggered touching 3+ match-lockable items within the match-lock duration.

The match-lockable items vary by chapter. In the kit preparation screen it is food, water, and first aid. There is no match-locking on the save yourself screen or the voting screen. **[TODO: Discuss with team if we should allow match-locking on the vote screen which would trigger an immediate end to voting.]** During the save others screen, the match-lockable items can only be food, water, and first aid as before, but only if those are the correct choice for the task.

- [x] Basic match-locking mechanism
- [x] Kit prep match-locking
- [ ] Save others match-locking

You can't really test this without the touchscreen unless you want to do some crazy timeout magic. I will test thoroughly here...


Dianna:
Finishes resolving #1078

Dianna:
Finishes resolving #1078 
Resolves #705

packages/2019-disaster-game/src/components/atoms/OrbManagerHelpers.js
```diff
@@ -134,3 +55,83 @@ export function incompleteOrbHandler(currentOrb, tick, index, orbConfig) {
...
+
+//     // also pass a delay for sequential animations
+//     orb.delay = Math.random() * 0.5;
+
+//     // store the data
+//     orbCollection[i] = orb;
+//   }
+
+//   return orbCollection;
+// }
```
jaronheard: Guessing the large comment block here is sticking around for a reason, but just in case...

Dianna: Yup. This PR is still a draft. Not ready for review yet.

Dianna:
Won't do

PR #1084, opened 2019-10-07
fix button element

paigewilliams:
Addresses #871

In the issue, @fstorr sugguest two different ways to solve the problem. I felt that switching it to a `<button>` was less of a work around to solving the problem. I also updates the styling so it matches previous iterations.

### DIFF
--- a/packages/component-library/src/CivicStoryCard/CivicStoryLink.js
+++ b/packages/component-library/src/CivicStoryCard/CivicStoryLink.js
@@ -4,7 +4,7 @@ import { Link } from "react-router";
 import { jsx, css } from "@emotion/core";
 
 const storyLinkClass = css`
-  & a {
+  & a, button {
     display: flex;
     flex-wrap: nowrap;
     align-items: center;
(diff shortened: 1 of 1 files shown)


packages/component-library/src/CivicStoryCard/CivicStoryLink.js
```diff
@@ -4,26 +4,41 @@ import { Link } from "react-router";
...
     align-items: center;
     justify-content: space-between;
     padding: 8px 12px;
     border-bottom: none;
     color: #000;
+    border: none;
+    background-color: #eee;
+    &:focus {
+      outline: 0;
+    }
```
jaronheard: If we're going to set outline to 0 here, we should have some alternative styles for focus -- read more here for why: http://outlinenone.com

#823 is a related open issue but needs a design. For now we could remove this reset and leave the default focus.

paigewilliams: Thanks for the feedback! My latest commit removed the focus styling.

PR #1096, opened 2019-10-13
Offline tile server support for Earthquake Heroes

DingoEatingFuzz:
This adds conditional support for using a custom tileserver api. 

It's implemented in a manner that is optional and non-breaking. After this merges, running the game locally or at http://disastergame.civicplatform.org/ will continue to work as it always has been.

However, if you run the game with `TILESERVER` set in your environment, this value will be appended to the index.html file via the express server and made accessible via a selector in `state/settings`.

Example for local use:

```sh
$ TILESERVER=http://localhost:3456/tiles yarn start
```

You can also see how this will be used [on site at OMSI over here](https://github.com/hackoregon/earthquake-heroes-offline/blob/master/docker-compose.yml).

### DIFF
--- a/packages/2019-disaster-game/src/constants/env.js
+++ b/packages/2019-disaster-game/src/constants/env.js
@@ -1,0 +1,18 @@
+// Static site configuration as set in the environment of the server
+// TODO: This is currently written under the assumption that this module
+// will be lazy-evaluated after the DOM is ready. This isn't necessarily
+// true and could lead to some bewildering debugging.
+const env = {};
+
+const configMetaEl = document.head.querySelector('[name="config/environment"]');
+if (configMetaEl) {
+  try {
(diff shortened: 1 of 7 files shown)


packages/2019-disaster-game/src/constants/env.js
```diff
@@ -0,0 +1,18 @@
+// Static site configuration as set in the environment of the server
+// TODO: This is currently written under the assumption that this module
+// will be lazy-evaluated after the DOM is ready. This isn't necessarily
+// true and could lead to some bewildering debugging.
```
jaronheard: Looks like this isn't quite addressed? If this isn't an issue, maybe remove the comment?

DingoEatingFuzz: It's not an issue _today_. My concern is that something will change in the future that will cause this to break. In the event that that happens, maybe the comment is still too little too late?

PR #1116, opened 2019-10-23
Vector Tiles Map

mendozaline:
This is a minimal implementation of a map that uses vector tiles.

It works by having the `BaseMap` pass its `Map` instance to the `VectorTilesMap` component. It then adds sources and layers to that object.

Also includes a standard story example and notes.


packages/component-library/stories/vectorTilesMap.notes.md
```diff
@@ -0,0 +1,46 @@
...
+```json
+{
+  "fill-color": "#DC4556",
+  "fill-opacity": 0.75
+}
+```
+
+- **layerPosition:** id of an existing layer to insert the new layer before
+  - This prop requires a string
+  - For example: `waterway-label`, `bridge-path`
```
jaronheard: Is there a way to pick a default layerPosition or make a recommendation for the user of the component as to what works with our BaseMap?

Also, I'm assuming that this is a layer from BaseMap that the new layer gets inserted before (above?), but it would be good to clarify that.

mendozaline:
I addressed most of the issues from above. 

Still need to determine the best default value for `layerPosition` prop. It has to be a layer that all `mapStyles` have, otherwise it will throw an error when adding a layer.

Issue #1121, opened 2019-10-28
Change orb interaction from hold to tap


Dianna:
OMSI request based on player research / feedback

PR #1123, opened 2019-10-28
Make orb selection instant per OMSI request

Dianna:
Can now click / tap / touch to select orb rather than needing to hold the orb. Gauge still shows and orb still flies! Ran through the whole game and worked like a dream

### DIFF
--- a/packages/2019-disaster-game/src/components/atoms/Orb.js
+++ b/packages/2019-disaster-game/src/components/atoms/Orb.js
@@ -36,7 +36,7 @@ const defaultState = {
   hasAnimated: false
 };
 
-const pressSuccessDuration = 1;
+const pressSuccessDuration = 0.5;
 
 export default class Orb extends PureComponent {
   constructor(props) {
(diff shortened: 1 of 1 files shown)


Dianna:
Resolves #1121

packages/2019-disaster-game/src/components/atoms/Orb.js
```diff
@@ -75,21 +75,16 @@ export default class Orb extends PureComponent {
...
-        clearTimeout(timer);
-      } else {
-        this.setState({ isActive: false });
-        clearTimeout(timer);
-      }
-    }, pressSuccessDuration * 1000);
+    const { isActive } = this.state;
+    if (isActive) {
+      this.setState({ isComplete: true });
+    }
```
DingoEatingFuzz: It looks like this changes the way the game works a bit. Now the item is captured immediately instead of after a second? 

Probably unimportant, but figured I'd bring it up.

Dianna: That's exactly the mechanics we want. Codes a little weird, but it works.

PR #1131, opened 2019-11-04
Remove button

Dianna:
Removes the button that isn't working. I think we'd need to figure out how to do this the _proper_ way that combines react-router and material-ui buttons


jaronheard:
@Dianna note that this link is external, so we don't need to use react-router.

This looks like it was using an `a` tag with `to` and it should have been using `a` with `href`.

PR #1146, opened 2019-11-12
Issue 1022

cosmolightfoot:
Fixed the filtering behavior to include OR logic within a filtering category, and AND logic between filtering categories. The level of filtration complexity jumped dramatically between a flat array of tags and comparing a nested object of tags. I tried my best to avoid unnecessary iterating but I would appreciate some feedback on my approach. I get the feeling that the CardList.js file is getting a little heavy / difficult to read.


packages/component-library/src/CardList/CardList.js
```diff
@@ -219,14 +230,44 @@ const CardList = ({ CardRegistry, tagsList, projects }) => {
...
+                          setActiveTagsByCategory({
+                            ...activeTagsByCategory,
+                            [category]: [
+                              ...activeTagsByCategory[category].filter(
+                                activeTag => activeTag !== topic
+                              )
+                            ]
+                          });
+                        }
                       }}
```
Dianna: This function should be broken out of the jsx

packages/component-library/stories/CardList.story.js
```diff
@@ -4,7 +4,21 @@ import { storiesOf } from "@storybook/react";
...
+  FilterSampleCardE,
+  FilterSampleCardF,
+  FilterSampleCardG,
+  FilterSampleCardH,
+  FilterSampleCardJ,
+  FilterSampleCardK,
+  FilterSampleCardL,
+  FilterSampleCardM,
+  FilterSampleCardN
+} from "./CivicCardSampleFilter";
```
Dianna: It would be better practice here to dynamically generate these cards since they're almost identical and then dynamically generate the `CardRegistryMock`. As much as possible we want to try and avoid writing code more than once. A good rule of thumb is 1-2 repeats, but it depends on the nature of the code. This, however, is a great candidate for a component generator function!

packages/component-library/stories/CivicCardSample/SampleCardFilterData.js
```diff
@@ -0,0 +1,45 @@
...
+    <CivicCard
+      cardMeta={sampleCardMetaWithTags}
+      data={sampleCardData}
+      isLoading={isLoading}
+      Layout={Layout}
+    />
+  );
+
+  SampleCard.tags = tags;
+  console.log("TAGS TAGS TAGS", SampleCard.tags);
```
Dianna: Remove this console.log

packages/component-library/stories/CardList.story.js
```diff
@@ -31,73 +17,13 @@ const CardRegistryMock = {
     Transit: 1,
     Demo: 1
   },
-  entries: [
-    {
+  entries: sampleFilterCardsWithTags.map(sampleCard => {
+    return {
       slug: "template-file-card",
```
Dianna: The slugs were different before, but now they're all `"template-file-card"`. This is just for a story, so it probably doesn't matter in this case, but if this were for production code, we'd want to make sure to dynamically generate this whole object that is returned so that the slug can be unique.

packages/component-library/src/CardList/CardList.js
```diff
@@ -40,88 +43,7 @@ const emptyState = css`
...
-    }
-  },
-  nested: {
-    paddingLeft: theme.spacing(4)
-  },
-  storyCard: {
-    height: "1000px"
-  }
-}));
+const useStyles = makeStyles(cardListStyling);
```
Dianna: Nice way of breaking out that code!

packages/component-library/src/CardList/CardList.js
```diff
@@ -20,6 +20,9 @@ import { ThemeProvider } from "@material-ui/styles";
 import Header from "../Header/Header";
 import BrandColors from "../_Themes/Brand/BrandColors";
 import ProjectCard from "./ProjectCard";
+import cardListStyling from "./cardListStyling";
+import deriveActiveTagsListsFromFilterStatus from "./deriveActiveTagsList";
+import filterCardsBasedOnFilterStatus from "./filterCardsByActiveTags";
```
Dianna: Excellent naming for these files / functions. It's super clear what they do

packages/component-library/src/CardList/CardList.js
```diff
@@ -158,25 +81,28 @@ const CardList = ({ CardRegistry, tagsList, projects }) => {
...
+  // flat array of active tags
+  const activeTagsList = deriveActiveTagsListsFromFilterStatus(
+    filterStatus,
+    categoryNames
+  )[0];
+  // arrays of active tags separated by category
+  const activeTagsByCategory = deriveActiveTagsListsFromFilterStatus(
+    filterStatus,
+    categoryNames
+  )[1];
```
Dianna: We should avoid calling a function with the same arguments, and therefore returning the same outputs, multiple times. This becomes especially important when dealing with large amounts of data because of performance implications. It still matters though because there will be a higher maintenance cost if the `deriveActiveTagsListsFromFilterStatus` is changed in some way. It's pretty small now, but it is an unnecessary complexity. I would refactor this code to look something like:

```javascript
const deriveActiveTagsListsFromFilterStatusResults = deriveActiveTagsListsFromFilterStatus(
  filterStatus,
  categoryNames
);
// flat array of active tags
const activeTagsList = deriveActiveTagsListsFromFilterStatusResults[0];
// arrays of active tags separated by category
const activeTagsByCategory = deriveActiveTagsListsFromFilterStatusResults[1];
```

packages/component-library/src/CardList/filterCardsByActiveTags.js
```diff
@@ -0,0 +1,60 @@
+/*
+  function compares the tags on each story card to see if they match the
+  selected filter checkboxes. For a card to be shown it must have ONE tag that
+  matches a checkbox IN EACH CATEGORY.
+
+  Example: If "Race", "Portland", and "Oregon" are active filters, cards will pass that contain
+  "Race" and either "Portland", or "Oregon".
+
+  Passing Results: ["Race", "Portland"], ["Race", "Oregon"], ["Race", "Portland", "Oregon", "Banana"]
```
Dianna: Great docs

packages/component-library/src/CardList/CardList.js
```diff
@@ -215,13 +156,16 @@ const CardList = ({ CardRegistry, tagsList, projects }) => {
...
                         if (showAllStories) setShowAllStories(false);
-                        setActiveTags({
-                          ...activeTags,
-                          [topic]: !activeTags[topic]
+                        setFilterStatus({
+                          ...filterStatus,
+                          [category]: {
+                            ...filterStatus[category],
+                            [topic]: !filterStatus[category][topic]
+                          }
```
Dianna: The whole `onChange` function should be abstracted. As much as possible, we want to avoid doing JS in JSX

Dianna:
Great refactoring here. I tested locally and it appears to be working great. There's just a couple of small changes I recommended above after which this will be good to merge 👍 

My software spidey senses are kicking in though and saying this code is still probably more complicated than it needs to be. That's being triggered by the amount of looping which I don't think _should_ be necessary. So if you are interested in doing a rethink / refactor session at some point let me know and I'd be down, but if you're feeling done with this feature, that's totally fine too. Regardless, we should merge this after those final fixes I recommended as it does improve the code and do the thing it's supposed to!

packages/2018/src/tags.js
```diff
@@ -1,5 +1,25 @@
...
+    "Race",
+    "Earthquake",
+    "Homeownership",
+    "Wealth",
+    "Redlining",
+    "HOLC",
+    "Permits",
+    "ADUs",
+    "Black",
+    "Gentrification"
```
Dianna: @jaronheard, do you want these extra tags included or removed?

packages/2018/src/tags.js
```diff
@@ -1,5 +1,25 @@
...
+  ],
+  locations: [
+    "Portland",
+    "Oregon",
+    "Nationwide",
+    "Your City",
+    "Tillamook County"
```
Dianna: Another new tag @jaronheard

packages/2018/src/tags.js
jaronheard: Remove Black, keep the rest

PR #1150, opened 2019-11-20
Feature/mapped howls

Dianna:
All the howls are now created once and stored as constants. I've played through a couple of times and they do indeed remain at a constant 40 howls which is the number of all sound files we use so that makes sense. YAY!


packages/2019-disaster-game/src/constants/sfx.js
```diff
@@ -1,63 +1,444 @@
...
+// Themes
+const attractorHowl = (() =>
+  new Howl({
+    src: [attractorSong],
+    autoplay: false,
+    loop: true,
+    volume: 0.25,
+    onfade: () => {
+      attractorHowl.stop();
+    }
```
DingoEatingFuzz: If you use a class method here instead you can avoid referencing the outer variable:

```js
onFade() {
  this.stop();
}
```

Not a necessary change, just something I noticed.

packages/2019-disaster-game/src/constants/sfx.js
```diff
@@ -1,63 +1,444 @@
...
-  loop: false
+  ...defaultHowl,
+  src: [badBasketball]
+});
+
+const badgeEarnedSFX = new Howl({
+  ...defaultHowl,
+  src: [badgeEarned],
+  volume: 0.25
+});
```
DingoEatingFuzz: If you're bored (lol) and looking for ways to tighten up the repetitiveness of this file, it might be nice to make a "factory" for these common Howl params.

Something like...

```js
const makeSound = (src, options = {}) => new Howl({ ...defaultHowl, ...options, src: [src] });

const badBasketballSFX = makeSound(badBasketball);
const badgeEarnedSFX = makeSound(badgeEarned, { volume: 0.25 });
```

PR #1165, opened 2019-12-09
Fix Geocoder event propagation

### LINKED ISSUE #1164: Interactions with the Geocoder in BaseMap propogate events to map
## Description

If you double click on the Geocoder (for example to select text), it will zoom the map.

### Steps to reproduce

Storybook example, double click the geocoder:
https://hackoregon.github.io/civic/?path=/story/component-lib-maps-base-map--example-with-geocoder

### Expected result

Should not cause any change to the underlying map

### Actual result

Zooms the map

### Environment

Production

mendozaline:
This resolves #1164 by moving the geocoder outside the `MapGL` component so that the search field no longer zooms the Base Map.


jaronheard:
Actually, that's being linked to from our OMSI exhibit, so let's fix it before merging.

PR #1176, opened 2019-12-18
Sandbox dashboard improvements

jaronheard:
This PR is mostly focused on improvements to the Sandbox dashboard, specifically the data details tab.

Improvements are as follows (by commit):
- Correctly enables and disables the data details tab
- Enlarges the polygon preview
- Opens that tab on the selection of a polygon with data
- Cleans up the code


Dianna:
I didn't rebuild the component library when I tested. The "bug" I mentioned does not exist. This looks good

PR #1182, opened 2020-01-16
Prepare for Netlify Deployment

jaronheard:
This PR sets up the CIVIC repo for deployment via Netlify. I've tested the Netlify deployment on [my fork](https://github.com/jaronheard/civic/).

There are four main changes:
- Add build-netlify script -- this has to use yarn workspaces instead of lerna to be compatible with Netlify
- Change http:// urls across the repo to https:// for API compatibility
- Add HTMLWebpackPlugin to generate an index file
- Add _redirects file for Netlify


packages/component-library/src/FooterNew/FooterNew.js
```diff
@@ -98,6 +98,9 @@ const Footer = props => (
       </div>
       <Logo css={logoStyle(props)} type="squareLogo" />
     </div>
+    <a href="https://www.netlify.com">
+      <img src="https://www.netlify.com/img/global/badges/netlify-color-bg.svg"/>
+    </a>
```
DingoEatingFuzz: Is this required with the nonprofit package?

jaronheard: There's a requirement to have a link for the open-source plan (https://www.netlify.com/legal/open-source-policy/). I'm going to tone it down a bit, though

jaronheard: Actually, removed this for now.

jaronheard:
@DingoEatingFuzz For now, the extent of this change is to allow for deploy previews. This would allow for us to transition away from ECS deployment, but you had identified a few key points around parity, security, and ✨the future✨ that were important considerations before committing to a full transition.

DingoEatingFuzz:
👍 Awesome, thank you for clarifying!

PR #1190, opened 2020-01-20
Update home page icons

jaronheard:
Updates the icons on the home page to be more consistent with each other


ghost (approved):
Good use of data viz colors

PR #1195, opened 2020-02-07
Feature/game ui updates

Dianna:
- Everything is now in Akkurat I also bolded the font so it still stands out. Italics were removed from the save yourself intro.
- Put underline in same light lime color as the interface under the active phase.
- Increased kit phase from 25 to 30 seconds.
- Shortened all task modal times by 2 seconds.
- Slowed down transition to badge slide by making it fade in and slowing the speed the solve screen pulls to the top of the screen.
- Made "1 pet" "2 people" type phrases display on same line by using non-breaking space character
- Added light lime colored rectangle with rounded edges around the copy of the summary screen
- Updated all copy adjustments based on E-Beth's doc update


jaronheard:
SO CLOSE! Just a couple of items before I would call it final.

image
The spacing on this button was driving me nuts. Can you add a margin-top of 1.5rem or so here? I think it should be more bold as well but this looks like the highest weight of Akkurat available.

The 3d effect on the text on the earthquake animation is also driving me nuts. Could we try doing it flat, or with the same depth as some of the other more subtle 3d effects?

PR #1200, opened 2020-03-07
Break up the component-library into smaller modules

jaronheard:
# Summary

This pull request breaks up the `component-library` into smaller packages, with the intention of replacing and deprecating the component library in favor of these smaller modules.

It leaves the current component-library architecture as the default for now, so all components are now duplicated across packages. The plan for moving to the new components will be detailed in an upcoming RFC.

## New packages

| Packages | Description | Dependencies | Dev Dependencies | Comments |
|----------|-------------|----------|----------|----------|
| ui-brand | Brand assets | ui-themes, utils | ui-docs |  |
| ui-cards | Story cards | ui-brand*, ui-core, ui-themes, utils | ui-charts | ui-brand: Logo used for loading spinner |
| ui-charts | Charts components | ui-brand*, ui-themes, utils |  | ui-brand: Logo used for loading spinner |
| ui-core | Common UI components | ui-themes, utils |  |  |
| ui-docs | Style guide and Storybook styles | ui-themes | ui-brand |  |
| ui-maps | Map components | ui-brand, ui-charts, ui-core, ui-themes, utils| |  |
| ui-themes | Themes and common styles | |  |  |
| utils | Utilities | |  |  |

## Next steps

- [x]  Ensure that the packages are broken out in a way that makes sense 
- [x]  Add read-me to all packages, adjust documentation
- [x]  Fix errors
- [x]  Make it possible to run Storybook for component-library (yarn new-storybook)
- [x]  Ensure that the correct version of Storybook is published to the web

## Future work
- [x]  Migration plan for component-library (https://github.com/hackoregon/civic/issues/1207)


jaronheard:
Per discussion with @DingoEatingFuzz, making the following adjustments:
- Removing `ui-labs`, in favor of developing within project packages and marking components experimental
- Removing `ui-platform` in favor of developing within project package
- Will check to see the size of ui-cards and ui-core? Could they bundled?

Dianna:
I can't provide the caliber of technical review that @DingoEatingFuzz can, so I'll focus on participatory design aspects. These are probably the kinds of things that I'd put in an RFC about the migration process:
* Should we transition all apps using the existing component library to this modular version and completely remove the old library? It would be ideal to preempt confusion about which library to use and I could see a newer dev thinking it's maybe easier to use the all-in-one-place-but-deprecated-and-enormous library. Alternatively, if migrating all apps using the deprecated library is hard, this could probably be accomplished with good documentation, but they'll need to be migrated over eventually anyways if we want them to benefit from component library updates (and keep our brand identity synced across apps)
* I'm a bit concerned about having READMEs that just contain links to storybook because contents and instructions may be different for different versions. Also, they aren't even linked to specific storybook areas — perhaps those structures should line up? The READMEs should probably contain at least basic pertinent info and then link out to storybook with a message like "see the latest version and how to use it on storybook"

jaronheard:
@Dianna, thanks for the review! In response to your bullets:
> Should we transition all apps using the existing component library to this modular version and completely remove the old library?

See #1207 for details on your first bullet, and let me know if you have further comments after reviewing that.

> I'm a bit concerned about the READMEs (sic)

Agreed, I'll flesh out the READMEs per your suggestion. I did the minimum viable README as part of my first PR with the expectation that I'd fill them out a bit once we knew how everything was shaking out, so thanks for the reminder.

packages/ui-brand/mocha.conf.js
```diff
@@ -0,0 +1,50 @@
...
+Object.keys(document.defaultView).forEach(property => {
+  if (typeof global[property] === "undefined") {
+    exposedProperties.push(property);
+    global[property] = document.defaultView[property];
+  }
+});
+
+// chai setup
+chai.use(sinonChai);
+chai.use(require("chai-enzyme")());
```
DingoEatingFuzz, line 50: All this mocha stuff isn't necessary if the package doesn't have tests. There's probably also a good way to share this conf—or parts of it—with other packages.

packages/ui-brand/package.json
```diff
@@ -0,0 +1,54 @@
+{
+  "name": "@hackoregon/ui-brand",
+  "sideEffects": false,
+  "version": "0.0.4",
+  "description": "Brand assets for CIVIC",
+  "main": "dist/index.js",
+  "modules": "es/index.js",
+  "scripts": {
+    "build:esm": "rimraf es && babel src --out-dir es --copy-files --no-comments",
+    "build:cjs": "rimraf dist && babel src --out-dir dist --copy-files --no-comments",
```
DingoEatingFuzz, line 10: It'd be nice to get to the point where we don't need to different build targets like this.

packages/ui-brand/package.json
```diff
@@ -0,0 +1,54 @@
...
+    "build": "BABEL_ENV=esm yarn run build:esm && BABEL_ENV=cjs yarn run build:cjs",
+    "configure": "yarn run build",
```
DingoEatingFuzz, line 12: I suspect this is copied from another package so I won't scrutinize but I think some of these scripts can be rethought and maybe trimmed back.

packages/ui-brand/scripts/start.js
```diff
@@ -0,0 +1,3 @@
+console.log(
+  "`yarn start` doesn't work here. Maybe you meant to run `yarn storybook` at project root?"
+);
```
DingoEatingFuzz, line 3: If these commands are just going to log a line, you can use an echo directly in the package.json file.

```json
"scripts": {
  "start": "echo '`yarn start` doesn\'t work here. Maybe you meant to run `yarn storybook` at the project root?'"
}
```
```

File: packages/ui-charts/assets/civic-logo-animated-invert.svg [hunk 5]
```diff
@@ -0,0 +1,26 @@
...
+		C75.1,87,63.9,87,56.9,80.2z"/>
+	<path class="st1" d="M264.5,80.2c-7-7-7-18.1,0-24.9c3.4-3.4,7.8-5.1,12.6-5.1c4.7,0,9.2,1.9,12.6,5.1l11.7-11.7
+		c-6.4-6.4-15.1-10-24.3-10s-17.8,3.6-24.3,10c-13.4,13.4-13.4,35.1,0,48.3c6.8,6.6,15.4,10,24.3,10s17.5-3.4,24.3-10l-11.7-11.7
+		C282.7,87,271.5,87,264.5,80.2z"/>
+	<path class="st2" d="M295.6,85.9c-8.8,9-25.4,11.5-37.4-0.4c-6.5-6.4-12.5-22.9,0.8-36.5"/>
+	<path class="st3" d="M328.1,122.7h-40.7L272,138l-15.3-15.3h-29H10.6V10.6L0.1,0.1v133h252.2l19.7,19.8l19.8-19.8h46.8L328.1,122.7
+		L328.1,122.7z"/>
+	<path class="st1" d="M0.1,0.1l10.5,10.5h317.5v112.1l10.5,10.3V0.1C338.6,0.1,0.1,0.1,0.1,0.1z"/>
+</g>
+</svg>
```
DingoEatingFuzz, line 26: Do these SVGs also need to be here if they are in `ui-brand`? Is the intent to keep `ui-charts` from being bloated by superfluous `ui-brand` content?

packages/ui-charts/assets/metadataQA.json
```diff
@@ -0,0 +1,448 @@
...
+    "whatdataprotectionstandardsdoesthisdatasetcomplywith": "don't know",
+    "doesthedatasetcontaininformationthatmightbeconsideredsensitiveorconfidential": "Yes",
+    "whatinformationdoesthedatasetcontainthatmightbeconsideredsensitiveorconfidential": "it's public, so not confidential, but includes names and addresses of owners of each lot in the region; and information about sale dates and prices. ",
+    "doesthedatasetcontaininformationthatmightbeconsideredinappropriateoroffensive": "No",
+    "emailaddress": "stephen.osserman@hackoregon.org",
+    "whatisyourteamwithintheorganization": "Housing",
+    "whatisyourorganization": "Hack Oregon - 2019 Project Season",
+    "whoareyou": "Stephen Osserman, Nick Kobel, Xander Blair"
+  }
+}
```
DingoEatingFuzz, line 448: Do these metadata json files need to be in both `ui-cards` and `ui-charts`?

packages/ui-core/src/ButtonNew/ButtonNew.js
```diff
@@ -0,0 +1,43 @@
+/** @jsx jsx */
+import { jsx } from "@emotion/core";
+import PropTypes from "prop-types";
+import MaterialButton from "@material-ui/core/Button";
+
+import { ThemeProvider } from "@material-ui/styles";
+import { MaterialTheme } from "@hackoregon/ui-themes";
```
DingoEatingFuzz, line 7: Have you run numbers yet on how large `ui-core` is now? I'm curious how much weight material, emotion, and our ui themes adds.

packages/ui-charts/assets/civic-logo-animated-invert.svg
jaronheard: Apparently there was a whole unused assets folder in `ui-charts` being useless...

packages/ui-charts/assets/metadataQA.json
jaronheard: Apparently there was a whole unused assets folder in `ui-charts` being useless...

packages/ui-core/src/ButtonNew/ButtonNew.js
jaronheard: Could this be right?
image

jaronheard: image

jaronheard:
> 1. There are some duplicated files across packages. Are those intentional or by accident?

The duplicated files that you identified was accidental. I've removed the extraneous asset folder that existed in `ui-charts`

> 2. What's the deal with the .toMove files? Is that part of a multistep rollout or something?

Yep, rather than update 2018 to use the new components now, I'm going to do that in a separate PR. The components that will live within 2018 are in a temporary `.toMove` folder, and will live within `src` in the future.

The tarball [size comparison](https://arve0.github.io/npm-download-size/):
image

image

packages/ui-core/src/ButtonNew/ButtonNew.js
jaronheard: This whole measuring things ad-hoc is a little problematic. I'd like to put in some automated checks for bundle size and performance.

packages/ui-brand/package.json
jaronheard: Will be addressed in #1208

packages/ui-brand/package.json
jaronheard: Will be addressed in #1208

packages/ui-brand/scripts/start.js
jaronheard: Will be addressed in #1208

jaronheard:
Ok, I think this is ready now -- the comments have either been resolved, or incorporated into #1207.

PR #1215, opened 2020-09-06
Use common babelrc file in ui-core

jaronheard:
For some reason `@hackoregon/ui-core` was using a different babel configuration than the other packages, and it breaks code-splitting. (Note, I was the reason 😅)

This makes it use the same babel configuration as the other `@hackoregon/ui-` packages.


ghost (approved):
LGTM

PR #1226, opened 2020-11-21
Add test and lint workflows

jaronheard:
Add test and lint workflows for packages that have changes since the last time that they were published


jaronheard:
Closing as an unfinished WIP

PR #1243, opened 2020-11-25
add kitchen sink

vincegonzalez:
Changes:
- added kitchen sink markup
- added base tw variables based on brand

Todo
- create variable conventions


packages/ui-core/src/KitchenSink/KitchenSink.stories.js
```diff
@@ -0,0 +1,20 @@
...
+
+export default {
+  title: "Unsorted/Components/KitchenSink",
+  parameters: { component: KitchenSink }
+};
+
+const Template = args => <KitchenSink {...args} />;
+
+export const Standard = Template.bind({});
+
```
jaronheard, line 16: Feel free to remove the entire Margaritte export, which will remove the story

jaronheard (approved):

packages/ui-core/src/KitchenSink/KitchenSink.stories.js
znmeb: Does the fact that you're adding the kitchen sink imply that you had everything but the kitchen sink already? :-)

jaronheard: @znmeb 😁 I'm pretty sure that's exactly what it means haha

jaronheard (approved):
The code here looks good so I'm going to merge into the tailwind branch to get the preview working

vincegonzalez:
- remove non prose elements form kitchen sink
- rename kitchen sink to civic prose
- apply brand typography to prose (except font size)
- add shades of colors to have alternatives

jaronheard:
The colors variants look really good @vincegonzalez!

I was wondering why the typography was so small in comparison to the [@tailwind/typography demo](https://tailwindcss-typography.netlify.app/), but then I realized that there are 5 different prose variants for different contexts. 

I was thinking that I would add those variants as stories on the CivicProse component, but now I'm thinking I may combine the CivicProse and Typography stories into the Typography story example.

jaronheard:
For the typography, it doesn't look like our Google Fonts are being loaded in the current Storybook publish, which I will also fix.

I'm also thinking that the thin font weight on the H1 looks out of place now that it's not a big size.

